PR #12, opened 2021-12-28
Hello, sbt-typelevel

armanbilge:
Executes https://github.com/typelevel/sbt-typelevel/discussions/11.

Early days still, but it's solid enough to self-host and publish snapshots:
https://s01.oss.sonatype.org/content/repositories/snapshots/com/armanbilge/sbt-typelevel-ci-release_2.12_1.0/0.4-a57db49-SNAPSHOT/

If you feel empowered, please give it a try.

https://index.scala-lang.org/armanbilge/sbt-typelevel/

Highlights:
1. Your choice of CI snapshots: off / "true" snapshot / "stable" snapshot.
2. I think I fixed the double release problem in CI if you push a tag with a commit. Haven't tested this yet.
3. Will eventually support both passphrase and non-passphrase protected keys. ~Support for the former is not quite ready yet,~ ~and haven't tested the latter~. Confirmed to work for both!

Any and all feedback much appreciated, although it would be cool to get this merged as-is and the feedback can come in the form of issues or ideally PRs. I don't claim to have any real knowledge of what I was doing here, just tried my best 🙃 

Update: published a stable hash release to maven. ~Last step: getting keys with passphrases to work.~ Done.


armanbilge:
Alright, definitely not perfect, but overall seems to be functional :)

```scala
addSbtPlugin("com.armanbilge" % "sbt-typelevel-ci-release" % "0.4-95ee54d")
enablePlugins(TypelevelCiReleasePlugin)
```

https://index.scala-lang.org/armanbilge/sbt-typelevel/

#### Notable features
1. Batteries-included: `sbt-typelevel-ci-release` sets you up for CI publishing MiMa-verified artifacts to sonatype under semantic versioning.
2. Flexible/modular: the "super-plugin" is composed of independent plugins for versioning, mima, scalac settings, gpg signing, publishing, CI, etc. Pick and choose to make your own personal super-plugin.
3. IMO the code is fairly straightforward. Each plugin has a well-defined responsibility. No weird inter-dependencies to keep track of and not even a `taskDyn`. I hope this will make contributing easy.
4. Support for GPG signing for keys with and without passphrases. This is determined automatically, no config required.
5. Opt-in snapshot publishing, either as "true" snapshots (default) or "Spiewak-style" hash-versioned releases.
6. No more double-publish mishaps when pushing a tagged commit. If a publish job is triggered by a commit (and not a tag) but there is a tag, it will not publish. The job triggered by the tag will publish normally.
7. Temurin 8 is the default JVM.
8. The new s01 sonatype host is the default. A single setting switches you back to the legacy host.

#### Full list of plugins
```scala
addSbtPlugin("com.armanbilge" % "sbt-typelevel-no-publish"          % "0.4-95ee54d")
addSbtPlugin("com.armanbilge" % "sbt-typelevel-settings"            % "0.4-95ee54d")
addSbtPlugin("com.armanbilge" % "sbt-typelevel-versioning"          % "0.4-95ee54d")
addSbtPlugin("com.armanbilge" % "sbt-typelevel-mima"                % "0.4-95ee54d")
addSbtPlugin("com.armanbilge" % "sbt-typelevel-sonatype"            % "0.4-95ee54d")
addSbtPlugin("com.armanbilge" % "sbt-typelevel-sonatype-ci-release" % "0.4-95ee54d")
addSbtPlugin("com.armanbilge" % "sbt-typelevel-ci-signing"          % "0.4-95ee54d")
addSbtPlugin("com.armanbilge" % "sbt-typelevel-ci"                  % "0.4-95ee54d")
addSbtPlugin("com.armanbilge" % "sbt-typelevel"                     % "0.4-95ee54d")
addSbtPlugin("com.armanbilge" % "sbt-typelevel-ci-release"          % "0.4-95ee54d")
```

.scalafmt.conf
```diff
@@ -0,0 +1,63 @@
+version = 3.3.0
```
rossabaker, line 1: Relatively unimportant, but how many of the non-defaults are necessary?  One of the virtues of a formatter is minimizing trivial differences in the wild.  The closer this hews to upstream defaults, generally the better.

(I make an exception on vertical alignment because of merges.  I don't claim to not be a hypocrite.)

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,14 @@
...
+
+We are committed to providing a friendly, safe and welcoming environment for all, regardless of level of experience, gender, gender identity and expression, sexual orientation, disability, personal appearance, body size, race, ethnicity, age, religion, nationality, or other such characteristics.
+
+Everyone is expected to follow the [Scala Code of Conduct] when discussing the project on the available communication channels.
+
+## Moderation
+
+Any questions, concerns, or moderation requests please contact a member of the project.
+
+- Ross A. Baker: [gitter](https://gitter.im/rossabaker) | [twitter](https://twitter.com/rossabaker) | [email](mailto:ross@rossabaker.com)
```
rossabaker, line 11: I'm on chat hiatus.

```suggestion
- Ross A. Baker: [twitter](https://twitter.com/rossabaker) | [email](mailto:ross@rossabaker.com)
```

build.sbt
```diff
@@ -1,57 +1,129 @@
...
-homepage := Some(url("http://typelevel.org/"))
-
-resolvers += Resolver.sonatypeRepo("releases")
-
-// This is both a plugin and a meta-plugin
-
-addSbtPlugin("com.typesafe.sbt" % "sbt-pgp" % "0.8.2")
+ThisBuild / tlBaseVersion := "0.4"
+// ThisBuild / organization := "org.typelevel"
+ThisBuild / organization := "com.armanbilge"
```
rossabaker, line 5: Is this a rights thing or a this-is-still-experimental thing?

ci-signing/build.sbt
```diff
@@ -0,0 +1,2 @@
+addSbtPlugin("com.codecommit" % "sbt-github-actions" % "0.14.2")
+addSbtPlugin("io.crashbox" % "sbt-gpg" % "0.2.1")
```
rossabaker, line 2: Is Scala Steward smart enough to find per-directory builds?

ci/build.sbt
```diff
@@ -0,0 +1,2 @@
+addSbtPlugin("com.codecommit" % "sbt-github-actions" % "0.14.2")
```
rossabaker, line 1: This version got repeated vs. `ci-signing`.

ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
```diff
@@ -0,0 +1,29 @@
...
+import com.typesafe.tools.mima.plugin.MimaPlugin
+
+object TypelevelCiPlugin extends AutoPlugin {
+
+  override def requires = GitHubActionsPlugin && GenerativePlugin && MimaPlugin
+  override def trigger = allRequirements
+
+  override def buildSettings =
+    addCommandAlias(
+      "ci",
```
rossabaker, line 16: Thought exercise: how would we make this open to extension for projects that run additional checks?  This has frustrated me in other plugins that use aliases.

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
```diff
@@ -0,0 +1,30 @@
...
+    lazy val tlFatalWarningsInCi = settingKey[Boolean](
+      "Convert compiler warnings into errors under CI builds (default: true)")
+  }
+
+  import autoImport._
+  import TypelevelSettingsPlugin.autoImport._
+  import GitHubActionsPlugin.autoImport._
+
+  override def globalSettings = Seq(
+    tlFatalWarningsInCi := true
```
rossabaker, line 23: I have never been particularly fond of this setting varying from local development.

ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
rossabaker: Oh, `replaceCommandAlias`?

kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
```diff
@@ -0,0 +1,27 @@
...
+import sbt.plugins.JvmPlugin
+
+object TypelevelKernelPlugin extends AutoPlugin {
+
+  override def requires = JvmPlugin
+  override def trigger = allRequirements
+
+  object autoImport {
+    lazy val tlIsScala3 = settingKey[Boolean]("True if building with Scala 3")
+    def replaceCommandAlias(name: String, contents: String): Seq[Setting[State => State]] =
```
rossabaker, line 13: Does this deserve a `tl` prefix since it's auto-imported?

kernel/src/main/scala/org/typelevel/sbt/kernel/V.scala
```diff
@@ -0,0 +1,63 @@
...
+      case (None, Some(patch)) => 1
+      case (Some(patch), None) => -1
+      case (Some(thisPatch), Some(thatPatch)) =>
+        val z = thisPatch.compare(thatPatch)
+        if (z != 0) return z
+        (this.prerelease, that.prerelease) match {
+          case (None, None) => 0
+          case (Some(_), None) => 1
+          case (None, Some(_)) => -1
+          case (Some(thisPrerelease), Some(thatPrerelease)) =>
```
rossabaker, line 39: `M9` vs. `M10`?  Splitting on word-number boundaries would do the right thing.

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
```diff
@@ -0,0 +1,82 @@
...
+      settingKey[String]("The base version for the series your project is in. e.g., 0.2, 3.5")
+    lazy val tlHashSnapshots =
+      settingKey[Boolean]("If true, a hash version implies this is a snapshot (default: true).")
+  }
+
+  import autoImport._
+
+  override def buildSettings: Seq[Setting[_]] = Seq(
+    versionScheme := Some("early-semver"),
+    tlHashSnapshots := true,
```
rossabaker, line 27: Is this going to add hash release to Maven Central by default?  I know mutable snapshots are problematic, but those non-monotonic releases cluttering Maven Central are gross in their own right.  I'd rather make it easy for people to cut milestones.

.scalafmt.conf
armanbilge: I don't have any attachment to these particular settings, I just like auto-formatting :) I copy this config around my projects. I think it originated in CE3.

(In case it's not clear, this is only for this repo—this plugin doesn't dictate anything about formatting.)

build.sbt
armanbilge: Rights.

ci-signing/build.sbt
armanbilge: I would think so, since it can also update files in the `project/` directory. This unusual setup is necessary so that the plugin can host itself (see the symlinks in `project/`).

ci/build.sbt
armanbilge: Probably possible to avoid. Just a bit trickier b/c of the self-hosting thing.

ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
armanbilge: Yep, got that here:
https://github.com/typelevel/sbt-typelevel/blob/f277ca92734b427e0db106d43fbdec8a6948eddb/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala#L11-L15

kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
armanbilge: auto-imports definitely need a prefix. The similar `isDotty` setting in sbt-spiewak is the reason http4s is not in the Scala 3 Community Build:
- https://github.com/djspiewak/sbt-spiewak/issues/83

kernel/src/main/scala/org/typelevel/sbt/kernel/V.scala
armanbilge: Thanks, this can definitely be improved.

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
armanbilge: Nope, the default is that snapshots go to the snapshots repo. However setting this `tlHashSnapshots` to false would achieve stable snapshots to Maven Central.

Btw, the versioning achieves monotonicity by also including the distance from the last non-prerelease tag in the same series, if available, before the hash.

armanbilge: Btw, this plugin published itself as a snapshot:
https://s01.oss.sonatype.org/content/repositories/snapshots/com/armanbilge/sbt-typelevel-ci-release_2.12_1.0/0.4-a57db49-SNAPSHOT/

I temporarily switched it to publish to Maven Central to test the signing machinery.

.scalafmt.conf
rossabaker: Yeah, let's not make this a distraction.  Anyway, I like slim ones in the spirit of having a more standardized dialect of Scala.

build.sbt
rossabaker: If you post on the Sonatype JIRA, I'll vouch for you.

ci/build.sbt
rossabaker: It will probably never get out of sync in practice, especially if Steward manages it.  It's ... probably fine.

ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
rossabaker: So much else is structured, and we manipulate aliases as semi-colon delimited strings.  But I guess we inherit that from SBT's model.  This is probably the best we can do.

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
armanbilge: True. But also, fatal warnings make development/experimentation extremely annoying.

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
rossabaker: I think a lot of confusion comes from imprecise vocab.  I think of "snapshots" as anything that has `SNAPSHOT` in the version and thus is treated as a mutable artifact and banished by common policy to the `snapshots` directory of the repository.  And a "release" is something that does not have `SNAPSHOT` in the name, but may be a "hash release" by having a hash in what you call the `V.prelease`.

So I think your setup makes sense, and the default does not offend me, but there's enough confusion around the name that we should document what happens when it's true and false.

armanbilge: 💯 I completely agree.

To add to the confusion, this decision is made at versioning time (hence, why it is in the versioning plugin). I.e. whether you append `-SNAPSHOT` to your version number. That then determines where it gets published.

When this setting is `true`, any version with a hash in it is deemed a `-SNAPSHOT`.

When this setting is `false`, `-SNAPSHOT` is appended to the version number _only_ if the git repo has uncommitted changes in it. Which should never happen in CI.

👍 to good docs with examples

build.sbt
armanbilge: @rossabaker https://issues.sonatype.org/browse/OSSRH-76765

Also, would be great to be added to this repo as well. Thanks!

rossabaker: Done.  I created an @typelevel/sbt-typelevel team and added @djspiewak and @ChristopherDavenport as well for prior build art.

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
rossabaker: I generally solve that by adjusting my scalacOptions in session while in that mode.  That way it's consistent with CI by default, and I can opt out while hacking.  But I won't die on this hill if other people like it.

armanbilge: Interesting. OTOH I generally solve this problem by actually paying attention to the warnings indicated in the IDE 😉

One compromise is providing a `prePR` command that emulates a CI compile.

rossabaker: I like the compromise, but I thought that was the intent of the `ci` alias?

armanbilge: I always thought of `ci` as what you want CI to do—run the entire test suite, run MiMa, generate docs, etc. etc. Meanwhile, `prePR` are "lighter-weight" checks you can do locally, e.g. formatting, compiling with fatal warnings, etc.

build.sbt
armanbilge: I'm on the team, but don't think I got any elevated permissions.

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
armanbilge: Hmm, this is trickier to implement than I thought.
```scala
addCommandAlias(
    "prePR",
    "; root/clean; set tlFatalWarnings := tlFatalWarningsInCi.value; +root/Test/compile")
```
The problem is, how do you reset `tlFatalWarnings` to the old setting after `prePR`? Not sure if this is possible, or my sbt-fu isn't good enough anyway. Someone smarter can chime in, otherwise we can shunt this for now.

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
```diff
@@ -0,0 +1,83 @@
...
+
+  override def requires = GitPlugin
+  override def trigger = allRequirements
+
+  object autoImport {
+    lazy val tlBaseVersion =
+      settingKey[String]("The base version for the series your project is in. e.g., 0.2, 3.5")
+    lazy val tlUntaggedAreSnapshots =
+      settingKey[Boolean](
+        "If true, an untagged commit is given a snapshot version, e.g. 0.4-00218f9-SNAPSHOT. If false, it is given a release version, e.g. 0.4-00218f9. (default: true)")
```
armanbilge, line 21: @rossabaker wdyt?

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
armanbilge: Renamed to `tlUntaggedAreSnapshots`, see below.

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
rossabaker: I like it.

kernel/src/main/scala/org/typelevel/sbt/kernel/V.scala
armanbilge: If you don't mind, can we shunt this into a follow-up PR? Ordering of pre-releases is not critical for MiMa, since pre-releases don't impact bincompat. Doing this right means accounting for `x.y.z-RCn` comes _before_ `x.y.z` but `x.y.z-c6d2fdf` comes _after_ `x.y.z`. So for this to actually work we'll have to make some assumptions about how people version their stuff.

armanbilge:
Let me know what else is needed here so we can land this. I know there's still a lot to iterate on, but at least for me it'll be much easier to keep track of in issues and separate PRs rather than trying to track all the comments in this one.

After this merges I'll open follow-up issues for all the unresolved comments.

armanbilge:
Nice, was able to publish a snapshot under `org.typelevel` :)

https://oss.sonatype.org/content/repositories/snapshots/org/typelevel/sbt-typelevel_2.12_1.0/0.4-df43b10-SNAPSHOT/

build.sbt
```diff
@@ -1,57 +1,135 @@
...
-  </developers>
+ThisBuild / tlBaseVersion := "0.4"
+ThisBuild / organization := "org.typelevel"
+ThisBuild / organizationName := "Typelevel"
+
+ThisBuild / scalaVersion := "2.12.15"
+
+enablePlugins(TypelevelCiReleasePlugin)
+ThisBuild / tlCiReleaseSnapshots := true
+ThisBuild / tlCiReleaseBranches := Seq("series/armanbilge")
```
armanbilge, line 11: I'll change this before we merge, but for now lets me publish from my branch.

```suggestion
ThisBuild / tlCiReleaseBranches := Seq("main")
```

armanbilge:
Darn, I'll have to rollback the `s01` default. https://github.com/typelevel/general/issues/114

build.sbt
rossabaker: Rights are fixed.  (See #13).

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
rossabaker: It could be done with a custom task, but it may not be worth the trouble.

kernel/src/main/scala/org/typelevel/sbt/kernel/V.scala
rossabaker: :+1: #14

rossabaker (approved):
I don't think there are any showstoppers here.  I'll give the snapshot a try on a lower-profile project.

armanbilge:
main just published as https://oss.sonatype.org/content/repositories/snapshots/org/typelevel/sbt-typelevel-ci-release_2.12_1.0/0.4-60549ca-SNAPSHOT/sbt-typelevel-ci-release-0.4-60549ca-SNAPSHOT.pom

PR #31, opened 2021-12-31
Extract out a GitHub plugin

armanbilge:
Yep, yet another plugin ...

This one automatically configures boilerplate settings (SCM info and Scala.js sourcemap URIs) assuming you are using GitHub. This was previously part of the settings plugin, but since it seems like many people don't want their scalac settings dictated to them I extracted out this less contentious part.


github/build.sbt
```diff
@@ -0,0 +1,2 @@
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.7.1")
```
armanbilge, line 1: We have to be careful with our versioning strategy here. I'm rolling this back from 1.8 because otherwise I can't use it with http4s-dom which is currently blocked by upstream issues with 1.8, see https://github.com/http4s/http4s-dom/pull/60#issuecomment-1001079167.

github/src/main/scala/org/typelevel/sbt/TypelevelGitHubPlugin.scala
```diff
@@ -0,0 +1,52 @@
...
+        s"scm:git:https://github.com/$user/$repo.git",
+        s"scm:git:git@github.com:$user/$repo.git"
+      )
+
+    val identifier = """([^\/]+?)"""
+    val GitHubHttps = s"https://github.com/$identifier/$identifier(?:\\.git)?".r
+    val GitHubGit = s"git://github.com:$identifier/$identifier(?:\\.git)?".r
+    val GitHubSsh = s"git@github.com:$identifier/$identifier(?:\\.git)?".r
+    Try {
+      val remote = List("git", "ls-remote", "--get-url", "origin").!!.trim()
```
rossabaker, line 42: Probably worth calling out this is how it works.  Not everybody calls origin `origin`.

rossabaker (approved):

github/src/main/scala/org/typelevel/sbt/TypelevelGitHubPlugin.scala
armanbilge: True. I was mostly concerned with what the CI runners do, since you want to get this right on publish.

Issue #16, opened 2021-12-30
Re-organize top-level-plugins


armanbilge:
Currently, the super-plugins are `sbt-typelevel` (which brings in everything but CI release) and `sbt-typelevel-ci-release` (which adds in CI release). This was following after sbt-spiewak, to accommodate projects that don't want to do CI releases.

Based on feedback from the [Discord discussion yesterday](https://discord.com/channels/632277896739946517/920140670394589224/925812804978614273), I think I got it wrong. My take-aways were that:
- CI release is the real bottleneck/headache for almost everyone
- besides that, many people don't want to be dictated scalac settings, etc.

However, there is also Ross's desire in https://github.com/typelevel/sbt-typelevel/discussions/11#discussioncomment-1884972 for a plugin:
> Shrinks the typical build.sbt (or extensions like sbt-http4s-org) to the point that updates to this plugin _keep that build fresh_.

So, my new proposal is:
1. `sbt-typelevel-ci-release`: a super-plugin including versioning+mima+signing+sonatype+ci-release, but no more.
2. `sbt-typelevel`: a super-super-plugin, that gets you all of the above, plus whatever else is needed to achieve Ross's vision, so scalac settings, even scalafmt, sbt-header, mdoc, what-have-you.

(2) will be _very_ opinionated of course, but I think that's perfectly fine for many projects. Anyone who wants more fine-grained-control can always drop down to `sbt-typelevel-ci-release` and then build up their plugin/settings stack as they see fit.

rossabaker:
Objective 2 is hard: sbt-explicit-dependencies has repeatedly saved me, whereas many people hate it.  Scalafmt is not too controversial anymore, but has naysayers.  Headers are somewhere between?  And should scalafix be enabled out of the box?  Should certain scalafix linters be run out of the box?  We could fight for months!  

But the payoff to Objective 2 is that we could get a typelevel.g8 that creates projects with familiar workflows to our 70-and-growing maintainers, with a minimal build.sbt so older projects continue to work like newer projects.  Most of our builds ain't special.

armanbilge:
Yeah, I don't think we need to make it hard, just make it clear it's not for everyone, and the really good stuff is in the ci-release plugin anyway. Basically all the Typelevel projects I contributed to use scalafmt and headers so I think those are fine. They all use mdoc too, but maybe that belongs in a site plugin that also makes it as easy as ci-release (and again, not for those who have opinions).

PR #32, opened 2021-12-31
Re-organize super plugins

### LINKED ISSUE #19: sbt-header support?
One thing missing from sbt-spiewak is sbt-header support.  Maybe that's out of scope, but it's the first migration roadblock I hit.

armanbilge:
Based on #31. Executes #16. Closes #19.

This makes sbt-typelevel-ci-release a standalone super-plugin that brings in all the versioning/mima/ci release stuff but nothing more. This is what many maintainers seem to want.

Then, sbt-typelevel itself becomes an opinionated super-plugin, including scalac settings, formatting, and headers, on top of sbt-typelevel-ci-release.

IMHO this is the right approach: sbt-typelevel will immediately bootstrap a new repo into a very familiar setup. A project can always "graduate" to the lower-level sbt-typelevel-ci-release with a custom plugin stack, if the maintainers desire that.


core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
```diff
@@ -17,23 +40,57 @@ object TypelevelPlugin extends AutoPlugin {
...
 
   override def globalSettings = Seq(
     tlFatalWarningsInCi := true
   )
 
   override def buildSettings = Seq(
+    organization := "org.typelevel",
+    organizationName := "Typelevel",
+    startYear := Some(java.time.YearMonth.now().getYear()),
+    licenses += "Apache-2.0" -> url("http://www.apache.org/licenses/"),
```
armanbilge, line 57: Thoughts? These defaults keep even more boilerplate out of `build.sbt` and are very easy to override anyway.

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
```diff
@@ -17,23 +40,57 @@ object TypelevelPlugin extends AutoPlugin {
...
+  import HeaderPlugin.autoImport._
 
   override def globalSettings = Seq(
     tlFatalWarningsInCi := true
   )
 
   override def buildSettings = Seq(
+    organization := "org.typelevel",
+    organizationName := "Typelevel",
+    startYear := Some(java.time.YearMonth.now().getYear()),
```
rossabaker, line 56: We've got about 364 days until this turns into a bug.

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
rossabaker: Yes.  It's nice if Typelevel projects pick the same.  This would need to be synced with any prospective g8 template.

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
```diff
@@ -17,23 +40,57 @@ object TypelevelPlugin extends AutoPlugin {
...
     }
+  ) ++ replaceCommandAlias(
+    "ci",
+    "; project /; headerCheckAll; scalafmtCheckAll; scalafmtSbtCheck; clean; test; mimaReportBinaryIssues"
   )
 
+  override def projectSettings = AutomateHeaderPlugin.projectSettings ++
+    Seq(
+      headerLicense := {
+        // the following was copied from sbt-header
```
rossabaker, line 74: Why?  What is it doing which that plugin didn't?

rossabaker (approved):
I just want something that includes the batteries.

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
armanbilge: That's a good question. The comment I didn't add above this was `// copied from sbt-spiewak`.

I suspect you are right, we probably don't need this. @djspiewak has a whole `endYear` thing that I didn't copy here for simplicity (unless we want that?).

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
armanbilge: :) we can also just put this into the g8 and updated it every year.

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
rossabaker: `endYear` is controversial.  My view is that in most jurisdictions, it only matters when the youngest contributor dies, and we avoid complexity and ritual diffs by omitting it.

PR #41, opened 2022-01-02
Improve CI release config

armanbilge:
Two changes here:
1. Add a `tlCiReleaseTags` option, default `true`. This is so you can disable releasing tags from CI, good for projects like CE that may want CI snapshots but not CI tags.
2. Merge the `tlCiReleaseSnapshots` and `tlCiReleaseBranches` into a single setting `tlCiReleaseSnapshotBranches`, where you directly specify the branches you want snapshots from. Default is empty, which causes no snapshots to be released.

### DIFF
--- a/sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
+++ b/sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
@@ -26,8 +26,8 @@ object TypelevelSonatypeCiReleasePlugin extends AutoPlugin {
   object autoImport {
     lazy val tlCiReleaseTags = settingKey[Boolean](
       "Controls whether or not v-prefixed tags should be released from CI (default true)")
-    lazy val tlCiReleaseSnapshotBranches = settingKey[Seq[String]](
-      "The branches in your repository to release snapshots from in CI (default: [])")
+    lazy val tlCiReleaseBranches = settingKey[Seq[String]](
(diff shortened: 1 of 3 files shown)


armanbilge:
Btw, I'm using "snapshots" in the loose sense here. Whether they are `-SNAPSHOTS` or not depends on your versioning scheme. Really, it's whether to release on every push—how it releases, is determined by the versioning scheme.

sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
```diff
@@ -24,10 +24,10 @@ import sbtghactions.GenerativePlugin.autoImport._
...
 
   object autoImport {
-    lazy val tlCiReleaseSnapshots = settingKey[Boolean](
-      "Controls whether or not snapshots should be released (default: false)")
-    lazy val tlCiReleaseBranches =
-      settingKey[Seq[String]]("The branches in your repository to release from (default: [])")
+    lazy val tlCiReleaseTags = settingKey[Boolean](
+      "Controls whether or not v-prefixed tags should be released from CI (default true)")
+    lazy val tlCiReleaseSnapshotBranches = settingKey[Seq[String]](
+      "The branches in your repository to release snapshots from in CI (default: [])")
```
armanbilge, line 30:
```suggestion
    lazy val tlCiReleaseBranches = settingKey[Seq[String]](
      "The branches in your repository to release from in CI on every push. Depending on your versioning scheme, they will be either snapshots or hash releases. (default: [])")
```

rossabaker (approved):

Issue #34, opened 2022-01-01
Support CI matrices split across JVM/JS


armanbilge:
By default sbt-gh-actions generates a separate job for each scala/java version. However, many projects also have separate jobs for JVM/JS. Some thoughts:

1. Should we do this by default? Personally I like it because it's not always easy to tell from combined jobs whether a test failure occurred in JVM or JS if they are running the same tests. However, it increases the build matrix by 3 jobs, each which has some constant overhead. The decrease in CI time definitely worth for larger projects, but not so much for smaller ones.
2. To do this properly needs more support from sbt-gh-actions, which I think doesn't distinguish between caches for JVM/JS and thus they end up overriding each other. Also might be complicated to implement.
3. I tried to set this up once and ran into problems with CI release, I think related to the caching thing in (2). So it should be checked if that works (I volunteer bobcats for this).

Linking to https://github.com/typelevel/sbt-typelevel/issues/18 b/c trying to upstream this change to sbt-gh-actions might take longer than absorbing it 🙃

rossabaker:
For smaller projects, I'm less annoyed by CI time.  I like a default that's good for the larger projects.

PR #43, opened 2022-01-04
Introduce `tlCrossRootProject`

armanbilge:
Fixes https://github.com/typelevel/sbt-typelevel/issues/34.

This introduces a `tlCrossRootProject` helper for creating the "root" aggregate project. It:
1. Generates `root`, `rootJVM`, `rootJS`, and `rootNative` projects (all set to no-publish)
2. Automatically sorts any aggregated projects and cross-projects into their appropriate roots (`root` itself collects everything)
3. Adds an axis to the CI matrix for platform, but only adds jobs for platforms which are represented in the build

I've demonstrated/tested `tlCrossRootProject` on this build, where it adds a single `ciJVM` job to the matrix since there are no JS or native project in this build. (Note that we now also have a `rootJS` and `rootNative` project that are empty, just a small nuisance I hope.)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -105,7 +105,7 @@ lazy val ci = project
   .enablePlugins(SbtPlugin)
   .settings(
     name := "sbt-typelevel-ci"
-  )
+  ).dependsOn(noPublish)
 
 lazy val ciRelease = project
   .in(file("ci-release"))
--- a/build.sbt
+++ b/build.sbt
@@ -15,22 +15,19 @@ ThisBuild / developers := List(
   tlGitHubDev("djspiewak", "Daniel Spiewak")
 )
 
-lazy val root = project
-  .in(file("."))
-  .enablePlugins(NoPublishPlugin)
-  .aggregate(
-    kernel,
-    noPublish,
-    settings,
-    github,
-    versioning,
-    mima,
-    sonatype,
-    ciSigning,
-    sonatypeCiRelease,
-    ci,
-    core,
-    ciRelease)
+lazy val root = tlCrossRootProject.aggregate(
+  kernel,
+  noPublish,
+  settings,
+  github,
+  versioning,
+  mima,
+  sonatype,
+  ciSigning,
+  sonatypeCiRelease,
+  ci,
+  core,
+  ciRelease)
 
 lazy val kernel = project
   .in(file("kernel"))
@@ -105,7 +102,8 @@ lazy val ci = project
   .enablePlugins(SbtPlugin)
   .settings(
     name := "sbt-typelevel-ci"
-  ).dependsOn(noPublish)
+  )
+  .dependsOn(noPublish)
 
 lazy val ciRelease = project
   .in(file("ci-release"))
--- a/build.sbt
+++ b/build.sbt
@@ -81,6 +81,7 @@ lazy val sonatype = project
   .settings(
     name := "sbt-typelevel-sonatype"
   )
+  .dependsOn(kernel)
 
 lazy val ciSigning = project
   .in(file("ci-signing"))
@@ -103,7 +104,7 @@ lazy val ci = project
   .settings(
     name := "sbt-typelevel-ci"
   )
-  .dependsOn(noPublish)
+  .dependsOn(noPublish, kernel)
 
 lazy val ciRelease = project
   .in(file("ci-release"))
--- a/ci/build.sbt
+++ b/ci/build.sbt
@@ -1,4 +1,4 @@
 addSbtPlugin("com.codecommit" % "sbt-github-actions" % "0.14.2")
 addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "1.0.1")
-addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject" % "1.1.0")
-addSbtPlugin("org.portable-scala" % "sbt-scala-native-crossproject" % "1.1.0")
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.0.0") // scala-steward:off
+addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.4.0") // scala-steward:off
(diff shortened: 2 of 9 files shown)


build.sbt
```diff
@@ -15,22 +15,19 @@ ThisBuild / developers := List(
...
+  settings,
+  github,
+  versioning,
+  mima,
+  sonatype,
+  ciSigning,
+  sonatypeCiRelease,
+  ci,
+  core,
+  ciRelease)
```
armanbilge, line 30: Here it is in action. Nothing fancy. Regardless of the name of the `val` it's assigned to, it always generates projects called `root`—I didn't want to fuss with macros.

ci/build.sbt
```diff
@@ -1,2 +1,4 @@
 addSbtPlugin("com.codecommit" % "sbt-github-actions" % "0.14.2")
 addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "1.0.1")
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.0.0") // scala-steward:off
+addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.4.0") // scala-steward:off
```
armanbilge, line 4: Since the CI plugin is trying not to be too prescriptive, and we don't use the JS/native plugins except to check for their presence on a project, these versions should be sufficient.

Hopefully this won't interfere with our other updates.

ci/src/main/scala/org/typelevel/sbt/CrossRootProject.scala
```diff
@@ -0,0 +1,192 @@
...
+      rootJS.disablePlugins(ps: _*),
+      rootNative.disablePlugins(ps: _*)
+    )
+
+  def aggregate(projects: CompositeProject*): CrossRootProject =
+    aggregateImpl(projects.flatMap(_.componentProjects): _*)
+
+  private def aggregateImpl(projects: Project*): CrossRootProject = {
+
+    val jsProjects = projects.filter(p => Plugins.satisfied(p.plugins, Set(ScalaJSPlugin)))
```
armanbilge, line 74: Hmm, these filters don't work like I ~expected~ wanted them to—`Plugins.satisfied` is too strict.

ci/src/main/scala/org/typelevel/sbt/CrossRootProject.scala
```diff
@@ -0,0 +1,191 @@
...
+  def aggregate(projects: CompositeProject*): CrossRootProject =
+    aggregateImpl(projects.flatMap(_.componentProjects): _*)
+
+  private def aggregateImpl(projects: Project*): CrossRootProject = {
+    val jsProjects =
+      projects.filter(_.plugins.toString.contains("org.scalajs.sbtplugin.ScalaJSPlugin"))
+
+    val nativeProjects =
+      projects.filter(
+        _.plugins.toString.contains("scala.scalanative.sbtplugin.ScalaNativePlugin"))
```
armanbilge, line 76: Too horrible?

ci/build.sbt
rossabaker: Does the Native plugin have have the non-breaking track record of Scala.js?  What is the risk of this dependency to all the people who don't aspire to support Native, which I think is still most Typelevel projects?

rossabaker (commented):
I like to minimize how much users need to think about cross-platform, but I'm worried that extra dependenare going to make it hard for the typical project that just wants an easy way to publish for the JVM.

ci/build.sbt
armanbilge: All fair points, I'm not familiar enough with native but it's still much more in development. In any case, I ended up removing it because my plugin detection idea wasn't working. See https://github.com/typelevel/sbt-typelevel/pull/43#discussion_r777820570.

armanbilge:
Yeah, that's a fair point. I agree we should leave dependencies to native out, but SJS I think is stable enough?

Alternatively, all of this could be moved to separate plugins actually: `sbt-typelevel-ci-cross` for `tlCrossRootProject` and one more super-plugin `sbt-typelevel-cross` on top of `sbt-typelevel`.

rossabaker:
I fear heading for 2^n plugins for n supported use cases.  What is the approximate burden on a JVM-only project if the Scala.js machinery is present?

armanbilge:
Just the Scala.js plugin on the classpath. They can ignore this `tlCrossRootProject` thing, it's completely opt-in.

kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
```diff
@@ -26,12 +26,15 @@ object TypelevelKernelPlugin extends AutoPlugin {
...
-    def replaceCommandAlias(name: String, contents: String): Seq[Setting[State => State]] =
+
+    def tlReplaceCommandAlias(name: String, contents: String): Seq[Setting[State => State]] =
       Seq(GlobalScope / onLoad ~= { (f: State => State) =>
         f andThen { s: State =>
           BasicCommands.addAlias(BasicCommands.removeAlias(s, name), name, contents)
         }
       })
+
+    implicit def tlCommandOps(commands: List[String]): TlCommandOps = new TlCommandOps(commands)
```
armanbilge, line 37: I know implicits aren't the most popular. Thoughts on this `mkCommand` syntax to help define the aliases?

armanbilge:
These latest changes try to DRY the command alias stuff, since we end up with many more of these to handle `ciJVM`, `ciJS`, etc.

kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
rossabaker: I don't think this one does enough to justify the magic.  If we frequently stubbed our toe quoting semi-colons, maybe, but we don't, and the helper doesn't fix that anyway.

armanbilge: Fair enough, I'll scrap it.

rossabaker (approved):
We need to try this on an application, where cross-platform doesn't tend to be needed, and make sure that none of this interferes.  But it seems reasonable.

armanbilge:
Agree on testing 👍 

If I've done it right, it shouldn't interfere because the `TypelevelCiCrossPlugin` and `TypelevelCi{JS,JVM,Native}Plugin` are all so-called `noTrigger` plugins, i.e. they will only be activated if enabled explicitly. And that only happens if a `tlCrossRootProject` is created. Otherwise, everything here should sit dormant.

armanbilge:
This PR solved https://github.com/djspiewak/sbt-github-actions/issues/3.

Issue #42, opened 2022-01-02
Still have race-conditions / double-release problem for simultaneous push+tag push


armanbilge:
This condition was supposed to prevent it:

https://github.com/typelevel/sbt-typelevel/blob/ac41c6a9e9b721f777a68dfdcd74ff851f61ed6e/.github/workflows/ci.yml#L128

However, it seems that both `github.ref` and `github.ref_type` are defined by the ref that caused the workflow to start (makes sense in retrospect). So they will never be out-of-sync, which is precisely the indicator we rely on to determine if a push-triggered workflow is trying to do a proper tag release and stop it. So I need to rethink this.

Issue #21, opened 2021-12-30
Continue offering non-CI/GHA versions of plugins


armanbilge:
This one comes from @bpholt, to avoid issues with sbt-project-matrix which doesn't play nice with sbt-gh-actions. The current setup caters perfectly fine for this use-case (e.g., the separate `sbt-typelevel-sonatype` and `sbt-typelevel-sonatype-ci-release` plugins. It's a minor tax on this repo, but much more friendly to downstream.

armanbilge:
I'm not sure how important this is anymore, thanks to some developments.
- I'm hopeful that https://github.com/typelevel/sbt-typelevel/pull/64 will fix the sbt-projectmatrix interop
- Even Rob was won over by sbt-gh-actions 😝

PR #45, opened 2022-01-04
Take 2: fix CI release/snapshot race condition

armanbilge:
Fixes https://github.com/typelevel/sbt-typelevel/issues/42.

For context, this is the "race condition" where if a commit is pushed to GH with a tag and CI release is enabled for both tags and snapshots, two publish jobs will start simultaneously and both will have the identical (tagged) version. The workaround is to first push the commit and wait for its (snapshot) publish job to complete, then push the tag separately to trigger the second publish job.

Fixing this problem is also important for projects that may want CI release for snapshots but don't want CI release for tags, such as CE—there needs to be a way for the snapshot publish job to ignore the tags.

The solution here definitely makes me feel a bit queasy: it changes our versioning algorithm depending on the presence/value of an `GITHUB_REF_TYPE` environment variable, that indicates we are:
1. running in GH actions
2. running in a job that was _not_ triggered by a tag push

When both these conditions are true, the versioning algorithm ignores any "v" tags and adopts a hash version.

One of the nice things about this approach is that we still get one snapshot per push, in addition to any proper release. My previous solution was attempting to disable the snapshot publish job completely.

Thoughts?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,5 +125,5 @@ jobs:
           echo "$PGP_PASSPHRASE" | gpg --pinentry-mode loopback --passphrase-fd 0 --import /tmp/signing-key.gpg
           (echo "$PGP_PASSPHRASE"; echo; echo) | gpg --command-fd 0 --pinentry-mode loopback --change-passphrase $(gpg --list-secret-keys --with-colons 2> /dev/null | grep '^sec:' | cut --delimiter ':' --fields 5 | tail -n 1)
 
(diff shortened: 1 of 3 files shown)


rossabaker (approved):
Is this the first time `versioning` turns GitHub-specific?  That's gross, but I don't yet see a better way.

armanbilge:
Yep, until now `versioning` didn't know about GitHub at all. I considered pushing this further down the plugin stack to once of the `ci` plugins that are GitHub-only. But considering https://github.com/typelevel/sbt-typelevel/issues/21 not everyone who does GitHub CI release is using our `ci` plugin with sbt-gh-actions. So this is the only way to make the fix broadly applicable.

PR #46, opened 2022-01-07
Add `tlCursedVersions` setting

armanbilge:
For excluding mucked up or otherwise unpublished versions from mima checks. We've all been there. (Name subject to bikeshed.)


rossabaker:
I usually just run my own filterNot on these, which can additionally target specific Scala versions.  I don't have an objection to it, but I don't know that the setting adds much here.

armanbilge:
That's a good point. I think the `filterNot` on `mimaPreviousArtifacts` is slightly more tedious because it's a set of module IDs and not versions. In any case I'm happy to let this sit and see if anyone really wants it.

armanbilge:
Maybe the setting we are actually missing is a `tlPreviousVersions`, that is auto-populated by our plugin and then can be introspected/modified by the user.

rossabaker:
In that world, is tlVersionIntroduced still useful?  I guess it helps people avoid filtering on an ordering of whatever semver representation is settled on.

armanbilge:
I've migrated a lot of builds and this hasn't come up. Happy to revisit if someone wants it.

Issue #61, opened 2022-01-09
Platform-cross-aware GHA steps


armanbilge:
When you define an sbt step in sbt-gh-actions, it automatically prefixes the step with e.g. `++2.13.7` to set the Scala version as defined in the matrix.

When a matrix is split on platform (JVM/JS/Native), it would be helpful to have a step prefixed with e.g. `project rootJVM` or `project rootJS` with whatever is defined in the matrix.

I think this is possible, something like this: just as we have axes for OSes, JVMs, and Scala versions in the matrix, we need an axis of "aggregate projects". Each aggregate project would select a slice of sub-projects in the build (e.g., JVM ones or JS ones) and scope all the sbt steps in the workflow to that project, just like they are scoped to a particular Scala version.

I think this might also fix sbt-gh-actions' interop with sbt-project-matrix, which creates one sub-project per Scala version and thus wreaks havoc.

Linking to https://github.com/typelevel/sbt-typelevel/issues/18.

Issue #18, opened 2021-12-30
Bring in sbt-github-actions


armanbilge:
https://github.com/djspiewak/sbt-github-actions

Another important one, that IMHO could do with a bus-factor upgrade. This one is much larger though.

armanbilge:
I temporarily did this in #62 but this is still open for discussion.

armanbilge:
I mean, this happened. There's a plan that it should become a separate plugin again in the future, but I'm in no rush to shave that yak.

Issue #60, opened 2022-01-09
Should the single `ci` command alias be split into multiple workflow steps?


armanbilge:
Spinning out of https://github.com/http4s/sbt-http4s-org/pull/120#discussion_r780729127.

This is a pattern I adopted from sbt-spiewak, but it obviously has tradeoffs.

Pros:
- a single command, that in theory you can run locally to emulate CI
- works well when split into `ciJVM`, `ciJS`, etc.
- doesn't have to restart sbt for each step

Cons:
- harder to see on what step a build fails
- difficult to progressively build up the commands in the alias, or remove stuff
- sometimes, restarting sbt is actually a good thing

armanbilge:
I think fixing https://github.com/typelevel/sbt-typelevel/issues/61 would make it feasible to split CI into separate steps while being friendly to platform-crossed-matrices.

Issue #20, opened 2021-12-30
Don't bring in new plugins and outsource some of the existing 


armanbilge:
Since I'm opening #17 and #18, it's only fair to consider the opposite argument. @ChristopherDavenport in particular encouraged that we delegate to his existing no-publish and mima plugins, and offered to donate them to typelevel org. sbt-tpolecat is another one we could outsource to instead of our own settings plugin.

My concerns with this are:
1. bus factor
2. try getting your head around how sbt-gh-actions is released by sbt-spiewak without a direct dependency :)

rossabaker:
The more external repositories this integrates, the more liability if they break.  Also the harder the next breaking SBT update is going to be.  It's the classic monorepo vs. polyrepo debate.  Separate repos win if the functionality needs a separate release cycle (no) or have clients that don't want sbt-typelevel (presently yes, but not if this is successful?).

This is also an argument against the most ambitious ideas of #16.

armanbilge:
> have clients that don't want sbt-typelevel (presently yes, but not if this is successful?)

It's cheating, but if development of sbt-gh-actions moved here, it could help with the success thing. I think the ci release plugin we have here is a natural next step:

1. a pure workflow generation plugin
2. a fill-in-the-blank plugin, i.e., you provide the CI and publish commands, plugin generates the workflows
3. a complete ci-release plugin, i.e. (2) with the publish step filled in, you just bring creds

(1) and (2) are the two parts of sbt-gh-actions today. I'd guess some of its users probably want (3) as well.

armanbilge:
The current state of things is okay. Navigating sbt-tpolecat as an independent dependency is scary enough for now!

PR #62, opened 2022-01-09
Bring in sbt-gh-actions

armanbilge:
Closes https://github.com/typelevel/sbt-typelevel/issues/18.

I'm not dead set on this: I would just like to be able to iterate quickly for now so I can fix https://github.com/typelevel/sbt-typelevel/issues/61 and https://github.com/typelevel/sbt-typelevel/issues/60. I've also abandoned the tests for now, but those should be followed-up on.

This would become the largest and most complex plugin currently in this repo. Is it worth it?

IMO the relevant questions are:
1. Does it make sense to bring sbt-gh-actions under the Typelevel org? (imho yes)
2. Following (1), where should it live? And how should it publish itself? (still on the fence about this, but a monorepo does simplify things)
3. Besides changing the artifact groupid and package name, should we make any other scope-related changes?

Linking to:
- https://github.com/typelevel/sbt-typelevel/issues/20.


rossabaker:
This functionality is one of the most important things in the ecosystem with a bus factor of one.  Not mine to donate, but I'd love to see it graduate to Typelevel and hopefully take some pressure off @djspiewak.

armanbilge:
I'd like to move forward with this, at least temporarily. Projects are already starting to adopt sbt-typelevel and some important changes are currently blocked by this PR.

I had a Discord discussion earlier this week with @djspiewak and IIUC he is 👍 about sbt-gh-actions graduating to Typelevel but needs more time to think about whether or not it should specifically join in this monorepo.

For the majority of Typelevel projects some churn in sbt-gh-actions going into 0.4.0 final will not be a problem.
1. AFAIK there are no triangle dependencies with other plugins using sbt-gh-actions.
2. Builds are relying on autoimport rather than direct imports of sbt-gh-actions, so a change in package name would not be source-breaking.

armanbilge:
Discussion can continue in https://github.com/typelevel/sbt-typelevel/issues/18.

PR #64, opened 2022-01-09
Introduce `sbtStepPreamble` to GHA plugin

armanbilge:
Based on https://github.com/typelevel/sbt-typelevel/pull/62. Fixes https://github.com/typelevel/sbt-typelevel/issues/61.

With all modesty 😁 I think this is a simple and very powerful idea.

Currently, when sbt-gh-actions generates a `WorkflowStep.Sbt`, it always prepends the user commands with `++${{ matrix.scala }}`. So, the steps run with the Scala version set in the matrix. This is fantastic when your matrix is crossed on Scala versions.

However, our matrices are now often crossed on other things as well that are controlled via sbt, such as JVM/JS, or in JS projects, the JS runtime used (Node.js, browsers, etc.).

This PR exposes these "preamble commands" as a new `githubWorkflowBuildSbtStepPreamble` that defaults to `++${{ matrix.scala }}`. Some useful adjustments of this setting may be to:
1. add e.g. `project ${{ matrix.project }}` command, to select a `rootJVM` or `rootJS` project
2. add e.g. `set Global / useJSEnv := ${{ matrix.jsenv }}`, to change the runtime used for JS tests
3. remove `++${{ matrix.scala }}` altogether, for projects using sbt-projectmatrix to manage their builds

In general, I think this introduces a lot of flexibility, that will make it easier to implement https://github.com/typelevel/sbt-typelevel/issues/60.


rossabaker (approved):
This seems reasonable.

armanbilge:
Since this depends on https://github.com/typelevel/sbt-typelevel/pull/62 I'll wait on approval there before merging.

Issue #15, opened 2021-12-30
`prePR` command—or another solution to fatal-warnings-in-CI


armanbilge:
Spinning out of https://github.com/typelevel/sbt-typelevel/pull/12#discussion_r776804761.

The underlying problem is we often want fatal warnings in CI, but not locally since it makes development/experimentation annoying. But then, we end up fighting CI :)

One proposal is we add a `prePR` command that emulates a CI compile, by cleaning and enabling fatal warnings (if that is indeed how CI is configured).

armanbilge:
Ooh, another good one for `prePR` is to check the workflow is up-to-date. I just mucked that up.

PR #70, opened 2022-01-15
Define `prePR` command alias

armanbilge:
Fixes https://github.com/typelevel/sbt-typelevel/issues/15.

It's not pretty or efficient and many of us will PR without running it first ... but at least when things go wrong in CI, this makes it easier to fix/replicate it locally.


armanbilge:
We can land this one too. Maybe just `prePR` for convenience sake/familiarity? I possibly went overboard on the `tl` prefixes...

rossabaker:
Prefix both `ci` and `prePR` or neither, I'd say.  I don't have a strong view which.

armanbilge:
`ci` is currently a thing of the past (I need to merge main into this branch). `prePR` is going in the opinionated plugin, so it's probably safe to be unprefixed.

PR #86, opened 2022-01-20
Don't generate headers automatically on compile

armanbilge:
Currently headers are automatically generated on compile. This is a bad default and leads to mishaps in CI like https://github.com/http4s/http4s/issues/5917.

Headers must now be manually generated with `headerCreateAll`. This is consistent with formatting, which is also invoked manually. All of this is covered in the existing `prePR` command, although perhaps we need a "lighter" alias that just does headers+formatting without clean-compiling like `prePR` does.


rossabaker (approved):
:+1: This has bitten at least twice now.   Worth an 0.4.1, I'd say.

PR #92, opened 2022-01-23
Keep publish secrets scoped to specific steps

armanbilge:
Currently, they are available in the entire workflow. This adds a new setting `githubWorkflowPublishEnv` that allows us to scope these secrets to only the publish job which should be the only place they are used.

Technically a breaking change if someone was relying on the old global scoping.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,11 +52,6 @@ jobs:
             ~/Library/Caches/Coursier/v1
           key: ${{ runner.os }}-sbt-cache-v2-${{ hashFiles('**/*.sbt') }}-${{ hashFiles('project/build.properties') }}
 
-      - if: env.TEST == 'itworked'
-        env:
-          TEST: itworked
-        run: exit
-
       - name: Check that workflows are up to date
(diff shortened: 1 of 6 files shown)


armanbilge:
Actually, we can do even better here and keep them scoped to the specific steps maybe.

rossabaker (approved):
This doesn't reduce the malicious threat vector much, but it reduces the stupid threat vector, and that's a good thing.

armanbilge:
Huh. I must have broke the workflows?

.github/workflows/ci.yml
```diff
@@ -132,17 +128,26 @@ jobs:
           rm targets.tar
 
       - name: Import signing key
-        if: env.PGP_SECRET != '' && env.PGP_PASSPHRASE == ''
+        if: secrets.PGP_SECRET != '' && secrets.PGP_PASSPHRASE == ''
```
armanbilge, line 131: Secrets can't be used in conditionals. I'm not sure if we can use env here if we only set it for this step (not sure what happens first).

armanbilge:
Not 100% convinced but let's see what happens on main.

Issue #81, opened 2022-01-18
Fail silently on artifact upload/download steps?


armanbilge:
These seem to flake occasionally and are not really critical, just shuffling caches to save recompilation during the release job. So it would probably be okay for them to fail silently.

An example in https://github.com/typelevel/jawn/runs/4859261390?check_suite_focus=true#step:14:1.

```
Run actions/upload-artifact@v2
With the provided path, there will be 1 file uploaded
Starting artifact upload
For more detailed logs during the artifact upload process, enable step-debugging: https://docs.github.com/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging#enabling-step-debug-logging
Artifact name is valid!
Container for artifact "target-ubuntu-latest-2.13.8-temurin@8" successfully created. Starting upload of file(s)
Uploaded /home/runner/work/jawn/jawn/targets.tar (81.2%) bytes 0:8388607
A 503 status code has been received, will attempt to retry the upload
Exponential backoff for retry #1. Waiting for 5499 milliseconds before continuing the upload at offset 8388608
Finished backoff for retry #1, continuing with upload
Error: Unexpected response. Unable to upload chunk to https://pipelines.actions.githubusercontent.com/dr4d5xdeYJVOyUhrZtz8KK0R6jc49DIbpbB6Fk82oUW560k70a/_apis/resources/Containers/12175527?itemPath=target-ubuntu-latest-2.13.8-temurin%408%2Ftargets.tar
##### Begin Diagnostic HTTP information #####
Status Code: 400
Status Message: Bad Request
Header Information: {
  "cache-control": "no-store,no-cache",
  "pragma": "no-cache",
  "transfer-encoding": "chunked",
  "content-type": "application/json; charset=utf-8",
  "strict-transport-security": "max-age=2592000",
  "x-tfs-processid": "d1684538-5dcb-4a3f-8745-537477d6a514",
  "activityid": "451083d9-565b-4d05-bc87-ee9e822f2860",
  "x-tfs-session": "451083d9-565b-4d05-bc87-ee9e822f2860",
  "x-vss-e2eid": "451083d9-565b-4d05-bc87-ee9e822f2860",
  "x-vss-senderdeploymentid": "02120909-3c00-63c2-b451-f147b8aff5a4",
  "x-frame-options": "SAMEORIGIN",
  "x-cache": "CONFIG_NOCACHE",
[...]

Issue #74, opened 2022-01-17
Make artifact upload/download step cross-aware


armanbilge:
For matrices with independent jobs for JVM/JS, both jobs are uploading their artifacts under the same name. So when the publish job downloads these artifacts, it is only getting half, and has to recompile the other half during the release step.

armanbilge:
This overwriting behavior just caused the fs2 site to go down. The fs2 workflow uses the artifacts to pass the generated site to the site deploy job, however the site only existed in the JVM artifacts. I've applied a workaround, but https://github.com/typelevel/sbt-typelevel/pull/98 is a permanent fix.

armanbilge:
This was a known issue:
- https://github.com/djspiewak/sbt-github-actions/issues/17

PR #98, opened 2022-01-23
Support custom matrices for artifact upload/download

armanbilge:
Fixes https://github.com/typelevel/sbt-typelevel/issues/74. In doing so, I hope it fixes https://github.com/typelevel/sbt-typelevel/issues/81 as well.

This is a fairly complex change, consisting of the following:

1. Use _all_ the keys in the matrix to determine the upload artifact id.
  - Currently they are always `$os-$scala-$java`. If the matrix is crossed on other things, such as JVM/JS, this leads to artifacts with duplicate ids being uploaded. I suspect this may be what causes the failures we've been observing in CI.
  - Now, they are named `$os-$java-$scala-$key1-$key2` etc. This assumes that sensible value names are being used in the matrix (idk what the rules are for artifact upload names).
2. Introduce a `githubWorkflowArtifactDownloadExtraKeys` setting (please help bikeshed name).
  - There are two kinds of keys wrt to the publishing step. Some keys (such as the Scala version) we must download artifacts for each value of that key. Other keys (such as OS or Java), we only need to download the artifacts for the primary value of that key.
  - This setting is a `Set[String]` of keys of the first kind above (i.e., for which we must download artifacts for each value). We use it to add the key `project` (with values `rootJVM`, `rootJS` and/or `rootNative`) when setting up the platform cross in the matrix.
  - With this information, the publish workflow "re-creates" the matrix as a sequence of steps for downloading artifacts, heeding any exclusions.

These changes are fairly internal but possibly deserve a minor version bump. In any case, before releasing I will test the snapshot on some relevant projects.


rossabaker:
Regarding the choice of Set, does the order of keys matter, or do they just need to be in the same arbitrary order in order to generate a unique ID?

armanbilge:
Just a consistent ordering. Maybe should be a `SortedSet` here then?

rossabaker (approved):
Okay, if consistent-but-arbitrary is the goal, and that makes sense, then I think it's fine.

PR #101, opened 2022-01-26
Add "if relevant" versions of mima/publish commands

armanbilge:
This PR defines `tlPublishIfRelevant`, `tlPublishLocalIfRelevant`, and `tlMimaReportBinaryIssuesIfRelevant` and redefines `tlRelease` in terms of them. These are [inspired by sbt-spiewak](https://github.com/djspiewak/sbt-spiewak/blob/d689a5be2f3dba2c335b2be072870287fda701b8/core/src/main/scala/sbtspiewak/SpiewakPlugin.scala#L87).

These commands allow us to run e.g. `+tlMimaReportBinaryIssuesIfRelevant` even where there are holes in the `crossScalaVersions` for various subprojects. The if-relevant command checks if the current Scala version in `ThisBuild` belongs to the `crossScalaVersions` for that project, and if not, simply no-ops the command (with an info message logged).

I thought with Scala Native finally coming to Scala 3 we would finally be beyond these holes, but then I encountered this with spark in https://github.com/typelevel/frameless/pull/600.

### DIFF
--- a/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
+++ b/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
@@ -27,6 +27,12 @@ object TypelevelKernelPlugin extends AutoPlugin {
   object autoImport {
     lazy val tlIsScala3 = settingKey[Boolean]("True if building with Scala 3")
 
+    lazy val tlPublishIfRelevant = taskKey[Unit](
+      "A wrapper around the `publish` task which checks to ensure the current scalaVersion is in crossScalaVersions")
+
+    lazy val tlPublishLocalIfRelevant = taskKey[Unit](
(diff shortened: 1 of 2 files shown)


rossabaker:
It always seemed to me that these steps could be trained to respect `skip` rather than introducing a new key, but I never got around to trying to fix that in sbt-spiewak.  I particularly don't understand why `publishIfRelevant` is necessary when that does support skip, though I suspect there's a good reason:

```
> inspect skip
[info] Task: Boolean
[info] Description:
[info] 	For tasks that support it (currently only 'compile', 'update', and 'publish'), setting skip to true will force the task to not to do its work.  This exact semantics may vary by task.
```

armanbilge:
Oh, that's interesting. I'll look into that.

armanbilge:
https://github.com/typelevel/sbt-typelevel/pull/117 seems simpler.

Issue #69, opened 2022-01-13
`prePR` bot


armanbilge:
Building on https://github.com/typelevel/sbt-typelevel/issues/15.

Assuming that we get our `prePR` command, I want a way to ask CI to run `prePR` for me and commit the changes.

I wonder if this can be accomplished (mostly) by spitting out an additional workflow. The only trick bit is figuring out how to conveniently trigger that workflow.

rossabaker:
As long as it's opt-in, it could be useful.

PR #111, opened 2022-01-28
Add `tlPrePrBotHook`

armanbilge:
Another prePR-style alias, but this one skips the clean compile and is intended only for bots. Since sbt-typelevel absorbs so many plugins (sbt-gh-actions, sbt-scalafmt, etc.) this command will provide a unified way for scala-steward to run all the necessary post-upgrade steps. Will also be useful for https://github.com/typelevel/sbt-typelevel/issues/69 and any other ideas we come up with that require sbt-typelevel-managed files.


rossabaker:
Is Scala Steward going to know about this?

armanbilge:
Yep, that's the plan. This way we can update what's in this hook—and downstream plugins can even add additional steps—without steward itself needing to be adjusted.

rossabaker:
So for example http4s could add its scalafix lints.  Yeah, this seems reasonable, if Scala Steward is going to integrate with it.

rossabaker (approved):

armanbilge:
The parallel PR to steward was merged :)
- https://github.com/scala-steward-org/scala-steward/pull/2492

PR #115, opened 2022-01-29
Automatically populate `apiURL` with javadoc.io

armanbilge:
Only if there's not already something set.

### DIFF
--- a/sonatype/src/main/scala/org/typelevel/sbt/TypelevelSonatypePlugin.scala
+++ b/sonatype/src/main/scala/org/typelevel/sbt/TypelevelSonatypePlugin.scala
@@ -20,10 +20,11 @@ import sbt._, Keys._
 import com.typesafe.tools.mima.plugin.MimaPlugin
 import xerial.sbt.Sonatype, Sonatype.autoImport._
 import TypelevelKernelPlugin.mkCommand
+import sbt.plugins.IvyPlugin
 
 object TypelevelSonatypePlugin extends AutoPlugin {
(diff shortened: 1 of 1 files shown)


sonatype/src/main/scala/org/typelevel/sbt/TypelevelSonatypePlugin.scala
```diff
@@ -35,17 +35,19 @@ object TypelevelSonatypePlugin extends AutoPlugin {
...
-          List(
-            "reload",
-            "project /",
-            "+mimaReportBinaryIssues",
-            "+publish",
-            "tlSonatypeBundleReleaseIfRelevant"))
-      )
+    Seq(
+      tlSonatypeUseLegacyHost := true,
+      autoAPIMappings := true
```
armanbilge, line 40: There's no reason this is controversial, right?

rossabaker: "This requires that dependencies have this information in its metadata" -- I don't know the consequences if they don't.  It just doesn't link, right?

rossabaker (approved):
I like it.

sonatype/src/main/scala/org/typelevel/sbt/TypelevelSonatypePlugin.scala
armanbilge: Yep, I assume so.

Issue #108, opened 2022-01-28
Integrate Unidoc with TypelevelSitePlugin


armanbilge:
So far I've been cheating with javadoc.io URLs but that only works well for single-module projects. Laika supports API docs, so I think that should be easy. The hard part is that although my preference is for sites that are updated on every push, scaladocs should definitely only be updated on tags.

rossabaker:
Why should scaladocs only be updated on tags?  http4s published on push for years.  (Counterpoint: this is why the repo is enormous on first clone.)

armanbilge:
It's because as soon as you add a new method that hasn't been released, the scaladocs are misleading (this happened recently in Cats Effect).

rossabaker:
I would think it's the same basic tradeoffs as the mdocs: you don't want to document unreleased features, but you might not want to do a patch release to fix a typo in the {mdocs, scaladocs}.

armanbilge:
That's fair. The difference in my eyes is that scaladocs change in ordinary development but mdocs change more deliberately, when we actually write docs 😅

The real solution is what CE is doing now, which is publishing docs from very push to a 3.3.x branch which is strictly forwards/backwards compatible.

armanbilge:
To give a concrete example: the ember h2 stuff would already be showing up in http4s scaladocs. But if somebody adds some mdocumentation for it, we can easily hold that PR until the right time to merge.

rossabaker:
Okay, that's a fair distinction.

PR #116, opened 2022-01-29
Integrate Unidoc with TypelevelSitePlugin

armanbilge:
Closes https://github.com/typelevel/sbt-typelevel/issues/108.

- adds a `TypelevelUnidocPlugin` that kicks-in if the `ScalaUnidocPlugin` is enabled on the docs project
- API docs are copied into the Laika site, but only if this is a release tag (not a snapshot nor a pre-release tag)
- deploying the site now keeps existing files, so API docs persist even if they are not updated
- `homepage` and `apiURL` settings are populated automatically

One bit of trickiness is that even with this plugin enabled, your API docs URL will 404 until you do a tag release. This makes sense—in theory, you'd have to go back to the tag for the last stable release and deploy from there.

A temporary workaround is to either manually upload docs to the appropriately directory or set `tlSiteApiUri` to e.g. a javadoc.io url. Or just tag a release :)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -146,7 +146,7 @@ lazy val site = project
   .settings(
     name := "sbt-typelevel-site"
   )
-  .dependsOn(kernel, githubActions, noPublish)
+  .dependsOn(kernel, github, githubActions, noPublish)
 
 lazy val docs = project
   .in(file("mdocs"))
--- a/build.sbt
+++ b/build.sbt
@@ -151,4 +151,10 @@ lazy val site = project
 lazy val docs = project
   .in(file("mdocs"))
   .enablePlugins(TypelevelSitePlugin)
+  .enablePlugins(ScalaUnidocPlugin)
   .settings(laikaConfig ~= { _.withRawContent })
+  .settings(
+    laikaIncludeAPI := true,
+    autoAPIMappings := true,
+    laikaGenerateAPI / mappings := (site/Compile / packageDoc / mappings).value
+  )
--- a/build.sbt
+++ b/build.sbt
@@ -152,9 +152,7 @@ lazy val docs = project
   .in(file("mdocs"))
   .enablePlugins(TypelevelSitePlugin)
   .enablePlugins(ScalaUnidocPlugin)
-  .settings(laikaConfig ~= { _.withRawContent })
   .settings(
-    laikaIncludeAPI := true,
-    autoAPIMappings := true,
-    laikaGenerateAPI / mappings := (site/Compile / packageDoc / mappings).value
+    tlSiteApiUri := Some(uri("api/org/typelevel/sbt/")),
+    laikaConfig ~= { _.withRawContent }
   )
--- a/site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
+++ b/site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
@@ -38,6 +38,8 @@ object TypelevelSitePlugin extends AutoPlugin {
       "0.4.4")
     lazy val tlSiteApiUrl = settingKey[Option[URL]]("URL to the API docs")
     lazy val tlSiteApiUri = settingKey[Option[URI]]("URI to the API docs")
+    lazy val tlSiteKeepFiles =
+      settingKey[Boolean]("Whether to keep existing files when deploying site (default: true)")
     lazy val tlSiteGenerate = settingKey[Seq[WorkflowStep]](
       "A sequence of workflow steps which generates the site (default: [Sbt(List(\"tlSite\"))])")
     lazy val tlSitePublish = settingKey[Seq[WorkflowStep]](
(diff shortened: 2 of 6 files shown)


site/src/main/scala/org/typelevel/sbt/TypelevelUnidocPlugin.scala
```diff
@@ -0,0 +1,55 @@
...
+          homepage.value.map(_.toURI.resolve(api).toURL)
+      }
+    }
+  )
+
+  override def projectSettings = Seq(
+    laikaIncludeAPI := {
+      // include the API docs, only if this is not a snapshot nor a pre-release
+      // so long as tlSiteKeepFiles is true, it won't delete existing API docs
+      !isSnapshot.value && V.unapply(version.value).exists(!_.isPrerelease)
```
armanbilge, line 50: I'd like to make this configurable in the future. Feeling a bit overwhelmed with settings—should API docs be updated for snapshot pushes? pre-releases? How about the `VERSION` variable injected into mdocs, should it be the last stable version, or the latest tagged version including pre-releases?

site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
```diff
@@ -128,15 +148,16 @@ object TypelevelSitePlugin extends AutoPlugin {
             .toAbsolutePath
             .relativize(((Laika / target).value / "site").toPath)
             .toString,
-          "publish_branch" -> "gh-pages"
+          "keep_files" -> tlSiteKeepFiles.value.toString
         ),
         name = Some("Publish site"),
         cond = {
-          val predicate = tlSitePublishBranch
-            .value // Either publish from branch or on tags, not both
```
armanbilge, line 136: When I first wrote this, I was afraid about race conditions when publishing the same commit twice, for a branch and for a tag. But I think it's fine.

build.sbt
```diff
@@ -146,9 +146,13 @@ lazy val site = project
...
-  .dependsOn(kernel, githubActions, noPublish)
+  .dependsOn(kernel, github, githubActions, noPublish)
 
 lazy val docs = project
   .in(file("mdocs"))
   .enablePlugins(TypelevelSitePlugin)
-  .settings(laikaConfig ~= { _.withRawContent })
+  .enablePlugins(ScalaUnidocPlugin)
+  .settings(
+    tlSiteApiUri := Some(uri("api/org/typelevel/sbt/")),
```
armanbilge, line 156: Overriding this so we are teleported to the right package in the docs. Seems hard to configure automatically, but it's only a nice-to-have, not critical.

rossabaker:
This deserves a more thoughtful review, but how does all this play with the idea of just pointing at javadoc.io?  We're publishing our own scaladoc to the site _and_ pointing at that?

armanbilge:
Pointing at javadoc.io works fine for single-module projects, but not currently for multi-module projects that want all their scaladocs nicely condensed. I did wonder if we should publish an additional "unidoc" artifact to sonatype, precisely for this purpose.

What this plugin does, if enabled, is publish scaladocs to gh-pages and update all the API urls to point there instead of the javadoc.io default in the other PR. Which reminds me, I need to make sure the URLs account for versioning...

armanbilge:
> I did wonder if we should publish an additional "unidoc" artifact to sonatype, precisely for this purpose.

Expanding on this: this would provide stable unidocs for _every_ published version, rather than a e.g. `3.x` that is updated every release. The downside imo is that we wouldn't be hosting docs at our own domain 🙃

rossabaker:
javadoc.io may go away someday, but doc jars published to Maven Central should be around as long as Scala is around.  That Unidoc artifact is an interesting idea.

A lot of the bloat in the http4s repo is from all the years of publishing docs on each commit.  A good solution that got away from that would be nifty.

armanbilge:
Cool, I'll looking into the unidoc artifact idea a bit. It could simplify things nicely.

> A lot of the bloat in the http4s repo is from all the years of publishing docs on each commit. A good solution that got away from that would be nifty.

Scaladocs in particular, or docs in general? The gh-pages deploy action we are using here only makes a commit if something changes (although, running mdoc does always change some non-determinstic outputs).

rossabaker:
Scaladocs in particular.  Those generate big diffs from commit to commit with all the indexes and what not.

armanbilge:
Btw publishing a unidoc artifact is very feasible. All you need is an project with the unidoc plugin enabled and this setting:

```scala
Compile / packageDoc / mappings := (ScalaUnidoc / packageDoc / mappings).value
```

I can pursue this strategy instead.

armanbilge:
As it stands, this PR is largely superseded by https://github.com/typelevel/sbt-typelevel/pull/119 and https://github.com/typelevel/sbt-typelevel/pull/124.

The changes unique to this PR the direct integration between unidoc and laika, i.e.:
1. Copying the API docs directly into the laika site and deploying them together.
2. Automatically adding an API doc button to the site that points to the right place.

I'd definitely like to find a way to solve (2), but I need to think more about how to do it. Is (1) still valuable in a world where unidocs are viewable from javadoc.io? It depends how much a project cares about hosting docs at their own domain.

armanbilge:
I have an idea to fix (2) in https://github.com/typelevel/sbt-typelevel/pull/119#discussion_r801864152. I'm going to punt (1) since folks seem to like the published unidoc concept.

With that, this PR has run its course.

PR #117, opened 2022-01-30
Integrate with `skip`

armanbilge:
A (better) alternative to https://github.com/typelevel/sbt-typelevel/pull/101. This was a good lead.

1. Automatically sets `skip := true` if `crossScalaVersions` for a project does not contain the current `scalaVersion`.
2. Makes `mimaReportBinaryIssues` skip-aware.

I'm going to a snapshot this branch to test it out in frameless. Update: seems to work.

### DIFF
--- a/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
+++ b/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
@@ -48,7 +48,7 @@ object TypelevelKernelPlugin extends AutoPlugin {
     skip := {
       skip.value || {
         val cross = crossScalaVersions.value
-        val ver = (ThisBuild / scalaVersion).value
+        val ver = (LocalRootProject / scalaVersion).value
         !cross.contains(ver)
       }
     }
--- a/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
+++ b/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
@@ -44,6 +44,16 @@ object TypelevelKernelPlugin extends AutoPlugin {
   override def buildSettings =
     addCommandAlias("tlReleaseLocal", mkCommand(List("reload", "project /", "+publishLocal")))
 
+  override def projectSettings = Seq(
+    skip := {
+      skip.value || {
+        val cross = crossScalaVersions.value
+        val ver = (ThisBuild / scalaVersion).value
+        !cross.contains(ver)
+      }
+    }
+  )
+
   private[sbt] def mkCommand(commands: List[String]): String = commands.mkString("; ", "; ", "")
 
 }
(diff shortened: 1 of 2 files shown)


kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
```diff
@@ -44,6 +44,16 @@ object TypelevelKernelPlugin extends AutoPlugin {
   override def buildSettings =
     addCommandAlias("tlReleaseLocal", mkCommand(List("reload", "project /", "+publishLocal")))
 
+  override def projectSettings = Seq(
+    skip := {
+      skip.value || {
+        val cross = crossScalaVersions.value
+        val ver = (ThisBuild / scalaVersion).value
```
armanbilge, line 51: `ThisBuild` doesn't seem to be the right way to access the `++scalaVersion`

mima/src/main/scala/org/typelevel/sbt/TypelevelMimaPlugin.scala
```diff
@@ -41,6 +41,11 @@ object TypelevelMimaPlugin extends AutoPlugin {
   )
 
   override def projectSettings = Seq[Setting[_]](
+    mimaReportBinaryIssues := {
+      if ((mimaReportBinaryIssues / skip).value)
+        ()
+      else mimaReportBinaryIssues.value
+    },
```
rossabaker, line 48: This looks like a good upstream candidate.

kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
```diff
@@ -44,6 +44,16 @@ object TypelevelKernelPlugin extends AutoPlugin {
...
+  override def projectSettings = Seq(
+    skip := {
+      skip.value || {
+        val cross = crossScalaVersions.value
+        val ver = (LocalRootProject / scalaVersion).value
+        !cross.contains(ver)
+      }
+    }
+  )
+
```
rossabaker, line 56: I'd say this is a good upstream candidate too, but maybe too many builds are depending on it not being this way. :shrug:

rossabaker (approved):
This is the way Things Ought to Work.

kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
armanbilge: I was wondering about this. Wouldn't this break the trick we use to back-publish kind projector?

rossabaker: Oh, poo.  I think it would.  We could counterhack that, but yeah.  I don't think this would fly upstream.

mima/src/main/scala/org/typelevel/sbt/TypelevelMimaPlugin.scala
armanbilge: - https://github.com/lightbend/mima/issues/684

PR #119, opened 2022-01-30
Create `TypelevelUnidocPlugin`

armanbilge:
This one is pretty simple. A plugin to publish a unidoc artifact to sonatype/maven accessible via javadoc.io, e.g.:

`https://www.javadoc.io/doc/org.typelevel/sbt-typelevel_2.12/0.4.4/`

AFAIK since javadoc.io can't render snapshots, you can't really update docs between releases.


build.sbt
```diff
@@ -148,6 +148,13 @@ lazy val site = project
   )
   .dependsOn(kernel, githubActions, noPublish)
 
+lazy val unidoc = project
+  .in(file("unidoc"))
+  .enablePlugins(TypelevelUnidocPlugin)
+  .settings(
+    name := "sbt-typelevel-unidoc"
+  )
```
armanbilge, line 156: This is how you set it up. Choice of "unidoc" for the module name is arbitrary—it would be good to establish a best practice.

armanbilge: Actually: this doesn't might not even need a separate module. In theory you could enable the `TypelevelUnidocPlugin` on your core module and it would replace its API docs artifact with the unidoc artifact. Not entirely sure if this would work—could be playing with fire.

sonatype/src/main/scala/org/typelevel/sbt/TypelevelSonatypePlugin.scala
```diff
@@ -60,18 +60,21 @@ object TypelevelSonatypePlugin extends AutoPlugin {
       else
         "s01.oss.sonatype.org"
     },
-    apiURL := {
-      val javadocio = CrossVersion(
+    apiURL := apiURL.value.orElse(javadocioUrl.value)
+  )
+
+  private[sbt] def javadocioUrl = Def.setting {
+    if (isSnapshot.value) None // javadoc.io doesn't support snapshots
```
armanbilge, line 67: - https://github.com/maxcellent/javadoc.io/issues/8

build.sbt
```diff
@@ -148,7 +149,21 @@ lazy val site = project
...
+
 lazy val docs = project
   .in(file("mdocs"))
   .enablePlugins(TypelevelSitePlugin)
-  .settings(laikaConfig ~= { _.withRawContent })
+  .settings(
+    laikaConfig ~= { _.withRawContent },
+    tlSiteApiModule := Some((unidoc / projectID).value),
+    tlSiteApiPackage := Some("org.typelevel.sbt")
```
armanbilge, line 168: This is a bit clunky.

armanbilge: I have an idea to set `tlSiteApiModule := Some((unidoc / projectID).value)` automatically.

build.sbt
```diff
@@ -148,7 +149,21 @@ lazy val site = project
   )
   .dependsOn(kernel, githubActions, noPublish)
 
+lazy val unidoc = project
+  .in(file("unidoc"))
+  .enablePlugins(TypelevelUnidocPlugin)
+  .settings(
+    // Doesn't conflict w/ core artifact since this one won't have the sbt cross version suffix
+    // javadoc.io doesn't recognize sbt artifacts, so we must publish like this
+    name := "sbt-typelevel",
```
armanbilge, line 158: In retrospect not sure if naming the artifact `sbt-typelevel` is a good idea. `sbt-typelevel-docs` or `sbt-typelevel-unidoc` etc. is much better.

armanbilge:
I got additional support for this idea in https://github.com/erikerlandson/coulomb/discussions/233#discussioncomment-2137410. So I'm going to move ahead with it for the next release :)

build.sbt
armanbilge: A bit reflection-y but gets the job done :)

https://github.com/typelevel/sbt-typelevel/blob/0e013b49670667689f6ca43003fbda66ec38a330/sonatype/src/main/scala/org/typelevel/sbt/TypelevelUnidocPlugin.scala#L36-L37

PR #133, opened 2022-02-02
Fix favicon attaching in the site plugin

### LINKED ISSUE #121: None site favicon using default config
At the moment, generated microsites (with default site config from sbt-typelevel-site) don't contain a favicon. I guess that's because of an issue with resolving external links for favicons in the Laika.
https://github.com/typelevel/sbt-typelevel/blob/c7e198803c3bb3251738ae000bffb8e9bdf17876/site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala#L85
I could advise a straightforward fix to store favicon in a particular project but it isn't so handy. However, one can do it as a temporary workaround (we don't know how long it could be to fix that issue). WDYT?

danicheg:
This fixes #121.
@armanbilge thanks for the insights for fixing that issue, but I'm slightly changed the initial plan. Here we store only the favicon in the plugin resources, not the whole template. I think it fits the initial issue accurately. Principle of least power, you know :) Also, I've published the plugin locally and checked that fix works for the `log4cats` microsite (and it does).


danicheg:
I think that the root problem is in Laika. And when next Laika's release comes we will have to remove our tricky workarounds (it would be great if Laika will have all features we need). Though, it may take a long time because we haven't ETA right now if I get it accurately. And if you have plans to bring a template to the site plugin feel free to discard this PR. I just was about fixing the particular issue with the least tricky things :)

armanbilge:
Yeah, I think we will eventually want to copy the template so we can customize it. The [http4s.org](https://http4s.org/versions) site has some nice things because it uses a custom template, e.g. projects in left side bar, footer on every page, etc.

I thought it would be equally tricky to copying the favicon file which is why I suggested it, but no worries :) thanks, I'm really sorry to have mislead you and I really appreciate the PR ❤️

armanbilge:
Superseded by https://github.com/typelevel/sbt-typelevel/pull/180. Daniel, thank you as always for your contributions and support!

Issue #14, opened 2021-12-30
Sort prereleases by number within them (e.g., M9 < M10)


rossabaker:
Context: https://github.com/typelevel/sbt-typelevel/pull/12/files/ddf26e78666fdce3953581dd99e7a49ea79d4111#diff-f515c6fb650430625373f4dee21cd53c79e9464b0ab728b8fa02b25ee2dd6f47

armanbilge:
It's hard/impossible to anticipate all pre-release versioning schemes, so my suggestion is we support:
1. the Ms and RCs, which are used by Scala itself and several typelevel projects
2. the output of `git describe`, which is `tag-distance-hash` (this is what the built-in versioning emulates).

armanbilge:
Found some prior art:
https://github.com/scalacenter/versions
https://github.com/sbt/librarymanagement/pull/239

armanbilge:
I bet steward does something smart with version sorting actually.

PR #135, opened 2022-02-03
Prereleases come before releases

rossabaker:
On http4s-fabric, there are three tags:
* v1.0.0-M30
* v1.0.0-M30.1 (trying to publish)
* v1.0.0 (unpublished)

M30.1 is failing trying to MiMa the non-existent 1.0.0, because it thinks 1.0.0-M30.1 > 1.0.0.  It should be filtered: a prerelease comes before the corresponding final release.

See https://github.com/http4s/http4s-fabric/issues/1

### DIFF
--- a/kernel/src/main/scala/org/typelevel/sbt/kernel/V.scala
+++ b/kernel/src/main/scala/org/typelevel/sbt/kernel/V.scala
@@ -50,8 +50,8 @@ private[sbt] final case class V(
         if (z != 0) return z
         (this.prerelease, that.prerelease) match {
           case (None, None) => 0
-          case (Some(_), None) => 1
-          case (None, Some(_)) => -1
+          case (Some(_), None) => -1
+          case (None, Some(_)) => 1
           case (Some(thisPrerelease), Some(thatPrerelease)) =>
             // TODO not great, but not everyone uses Ms and RCs
             thisPrerelease.compare(thatPrerelease)


armanbilge:
The problem is there are two kinds of prereleases.

> Doing this right means accounting for `x.y.z-RCn` comes _before_ `x.y.z` but `x.y.z-c6d2fdf` comes _after_ `x.y.z`.

So, to get this "right" we have to account for both situations: making the change here would mean that `x.y.z-c6d2fdf` is not checked against `x.y.z`.

However, in practice we can shunt this issue because previous MiMa artifacts determined from previous tags, and there should never be a `v1.0.0` tag when the current version is `1.0.0-M30`. Is that the situation in fabric?

rossabaker:
Oh, I didn't think about snapshots.  I think Maven has special handling for "alpha", "beta", "RC", and "M".  Alternatively, it could be handled by looking for 7-hexdigit prereleases.  Or, as noted, it could be handled just by not pushing tags too early.

armanbilge:
I would like to solve this :) I looked into Coursier and it has a very good [`Version`](https://github.com/coursier/coursier/blob/94680bfc1a15bc06522a9cd5217be7f6b0658301/modules/core/shared/src/main/scala/coursier/core/Version.scala) model that seems to sort these right. So I think we can switch to that.

Closing for now since this is tracked by https://github.com/typelevel/sbt-typelevel/issues/14.

Issue #105, opened 2022-01-26
Fail if tag is older than base version


armanbilge:
As long as you have binary-breaking changes on a branch, then tagging a non-breaking version number will fail the bincompat checks.

However, currently in sbt-typelevel the series/0.4 branch and main (base version 0.5) are actually binary-compatible, but semantically incompatible. So accidentally tagging 0.4.x on main would succeed in publishing, which is not what we want.

rossabaker:
I just commented somewhere else, but I'd also like to be scolded if the base version is older than the latest tag on the branch.  I've seen MiMa failures at publish time because the new tag is the first time MiMa runs on new versions.

armanbilge:
Yes, I replied there too 😉  that is already supported: if the base version is older than the latest tag, it will fail to start sbt.

armanbilge:
@osleonard will work on this :)

armanbilge:
@osleonard btw let me know if you need any more pointers, since this is a bug I'd really love to get this into the next release :)

Here's where do the check in the opposite direction, if the baseVersion is older than the latest tag:
https://github.com/typelevel/sbt-typelevel/blob/58ad2e745bf4cd6122e1c88a2d4f3610876fab96/versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala#L76-L83

PR #136, opened 2022-02-03
#105

osleonard:
Fail if tag is older than base version #105

### DIFF
--- a/versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
+++ b/versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
@@ -90,7 +90,7 @@ object TypelevelVersioningPlugin extends AutoPlugin {
             sys.error(s"verion $version not formatted properly.")
           case Some(value) =>
             if (!(value.isSameSeries(baseV) || value >= baseV))
-              sys.error(s"Your latest tag $version cannot be less than tlBaseVersion $baseV")
+              sys.error(s"Your current version $version cannot be less than tlBaseVersion $baseV")
             else
               // Looks for the distance to latest release in this series
               latestInSeries.foreach { latestInSeries =>
--- a/versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
+++ b/versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
@@ -73,27 +73,27 @@ object TypelevelVersioningPlugin extends AutoPlugin {
           .previousReleases(true)
           .filterNot(_.isPrerelease) // TODO Ordering of pre-releases is arbitrary
           .headOption
+          .flatMap { previous =>
+            if (previous > baseV)
+              sys.error(s"Your tlBaseVersion $baseV is behind the latest tag $previous")
+            else if (baseV.isSameSeries(previous))
+              Some(previous)
+            else
+              None
+          }
 
         // version here is the prefix used further to build a final version number
         var version = latestInSeries.fold(tlBaseVersion.value)(_.toString)
 
-        latestInSeries.flatMap { previous =>
-          if (previous > baseV)
-            sys.error(s"Your tlBaseVersion $baseV is behind the latest tag $previous")
-          else if (version < baseV.toString)
-            sys.error(s"Your latest tag $version cannot be less than tlBaseVersion $baseV")
-          else if (baseV.isSameSeries(previous))
-            Some(previous)
-          else
-            None
-        }
-
-        // Looks for the distance to latest release in this series
-        latestInSeries.foreach { latestInSeries =>
-          Try(s"git describe --tags --match v$latestInSeries".!!.trim)
-            .collect { case Description(distance) => distance }
-            .foreach { distance => version += s"-$distance" }
-        }
+        if (version < baseV.toString)
+          sys.error(s"Your latest tag $version cannot be less than tlBaseVersion $baseV")
+        else
+          // Looks for the distance to latest release in this series
+          latestInSeries.foreach { latestInSeries =>
+            Try(s"git describe --tags --match v$latestInSeries".!!.trim)
+              .collect { case Description(distance) => distance }
+              .foreach { distance => version += s"-$distance" }
+          }
 
         git.gitHeadCommit.value.foreach { sha => version += s"-${sha.take(7)}" }
         version
--- a/versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
+++ b/versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
@@ -73,13 +73,15 @@ object TypelevelVersioningPlugin extends AutoPlugin {
           .previousReleases(true)
           .filterNot(_.isPrerelease) // TODO Ordering of pre-releases is arbitrary
           .headOption
-          .flatMap { previous =>
-            if (previous > baseV)
-              sys.error(s"Your tlBaseVersion $baseV is behind the latest tag $previous")
-            else if (!baseV.isSameSeries(previous))
-              sys.error(s"Your tlBaseVersion $baseV has to be in the same series with latest tag $previous")
+          .flatMap { currentTag =>
+            if (currentTag > baseV)
+              sys.error(s"Your tlBaseVersion $baseV is behind the latest tag $currentTag")
+            else if (currentTag < baseV)
+              sys.error(s"Your latest tag $currentTag cannot be less than tlBaseVersion $baseV")
+            else if (baseV.isSameSeries(currentTag))
+              Some(currentTag)
             else
-              Some(previous)
+              None
           }
 
         var version = latestInSeries.fold(tlBaseVersion.value)(_.toString)
--- a/versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
+++ b/versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
@@ -76,10 +76,10 @@ object TypelevelVersioningPlugin extends AutoPlugin {
           .flatMap { previous =>
             if (previous > baseV)
               sys.error(s"Your tlBaseVersion $baseV is behind the latest tag $previous")
-            else if (baseV.isSameSeries(previous))
-              Some(previous)
+            else if (!baseV.isSameSeries(previous))
+              sys.error(s"Your tlBaseVersion $baseV has to be same with latest tag $previous")
             else
-              None
+              Some(previous)
           }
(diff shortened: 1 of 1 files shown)


versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
```diff
@@ -76,10 +76,10 @@ object TypelevelVersioningPlugin extends AutoPlugin {
           .flatMap { previous =>
             if (previous > baseV)
               sys.error(s"Your tlBaseVersion $baseV is behind the latest tag $previous")
-            else if (baseV.isSameSeries(previous))
-              Some(previous)
+            else if (!baseV.isSameSeries(previous))
+              sys.error(s"Your tlBaseVersion $baseV has to be in the same series with latest tag $previous")
```
armanbilge, line 80: I'm not sure if this is what we want to check 🤔  IIUC this is checking if the base version is not in the same series `x.y` as the previous release, to throw an error. For example, if the base version is `0.5` and the previous release was `v0.4.3`. Actually, this is exactly the situation on the main branch of sbt-typelevel :)

AFAICT there's nothing wrong with this situation: it is normal for the base version to be ahead of the version of the last release, because it indicates you have started working on the next version of your project. You can test this by bumping the base version in `build.sbt` to 0.5 and reloading sbt.

I think this is why when your PR originally targeted the main branch it was failing:
https://github.com/typelevel/sbt-typelevel/runs/5047968979?check_suite_focus=true#step:5:47

An example of the situation that we do want to prevent is if the current version is `0.4.4` but the base version is `0.5`. The reason this is bad is because if the developed bumped the base version to `0.5` that must mean that their project is no longer compatible with the `0.4` series. In which case, if the current version is `0.4.4` then there is something wrong.

Let me know if that makes sense :)

osleonard: Makes sense :)

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
```diff
@@ -73,17 +73,21 @@ object TypelevelVersioningPlugin extends AutoPlugin {
...
           .filterNot(_.isPrerelease) // TODO Ordering of pre-releases is arbitrary
           .headOption
-          .flatMap { previous =>
-            if (previous > baseV)
-              sys.error(s"Your tlBaseVersion $baseV is behind the latest tag $previous")
-            else if (baseV.isSameSeries(previous))
-              Some(previous)
-            else
-              None
-          }
```
armanbilge, line 83: I'm a little confused, why did this code get removed?

osleonard: Move it further `version`  is derived from the optional value in `latestInSeries`.which make it convenient / easier to compare the latest and previous tag at once

armanbilge: I could be confused, but it seems like it changes the semantics. Because in the old code, the result of this `flatMap` is stored in the variable `latestInSeries`.

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
```diff
@@ -73,17 +73,21 @@ object TypelevelVersioningPlugin extends AutoPlugin {
...
 
+        // version here is the prefix used further to build a final version number
         var version = latestInSeries.fold(tlBaseVersion.value)(_.toString)
 
+        latestInSeries.flatMap { previous =>
```
armanbilge, line 80: However, in the new code the result of the `flatMap` isn't stored anywhere.

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
osleonard: You are right the value of the flatmap is lost, updated the code

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
```diff
@@ -82,14 +82,18 @@ object TypelevelVersioningPlugin extends AutoPlugin {
...
+        // version here is the prefix used further to build a final version number
         var version = latestInSeries.fold(tlBaseVersion.value)(_.toString)
 
-        // Looks for the distance to latest release in this series
-        latestInSeries.foreach { latestInSeries =>
-          Try(s"git describe --tags --match v$latestInSeries".!!.trim)
-            .collect { case Description(distance) => distance }
-            .foreach { distance => version += s"-$distance" }
-        }
+        if (version < baseV.toString)
```
armanbilge, line 88: Comparing versions by comparing `String` makes me nervous 😅 does this do what we need? Esp. since `version` is in the `x.y.z` form and `baseV` is in the `x.y` form.

armanbilge: Btw, the `V` model has some comparison capability.
https://github.com/typelevel/sbt-typelevel/blob/5fff779cd21c5c2e81eb667d9bfb9b0f42b97cc6/kernel/src/main/scala/org/typelevel/sbt/kernel/V.scala#L21-L26

osleonard: I totally forgot about that till I check the values in repl

two options on how I see it working 

- version.take(3) < baseV.toString that would give us the version in `x.y` form or
- we call apply method of V with the string and then pattern match on the value we get back.

WDYT ?

armanbilge: `version.take(3)`: what if `version` is `1.23`?

I like option 2 :) `V` is supposed to help working with versions strings.

One thing to be careful of: `V("x.y") > V("x.y.z")` for all `z`. This may not have been a good idea, but I originally designed it this way so the `baseVersion` is always in the "future". Actually I'm not sure if this is used anywhere anymore.

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
```diff
@@ -82,17 +82,25 @@ object TypelevelVersioningPlugin extends AutoPlugin {
...
+              sys.error(s"Your latest tag $version cannot be less than tlBaseVersion $baseV")
+            else
+              // Looks for the distance to latest release in this series
+              latestInSeries.foreach { latestInSeries =>
+                Try(s"git describe --tags --match v$latestInSeries".!!.trim)
+                  .collect { case Description(distance) => distance }
+                  .foreach { distance => version += s"-$distance" }
+              }
+            git.gitHeadCommit.value.foreach { sha => version += s"-${sha.take(7)}" }
+            version
```
armanbilge, line 102: Since this code is imperative anyway, do you mind if we leave these steps outside/after the `match` statement?

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
```diff
@@ -82,17 +82,25 @@ object TypelevelVersioningPlugin extends AutoPlugin {
...
-        latestInSeries.foreach { latestInSeries =>
-          Try(s"git describe --tags --match v$latestInSeries".!!.trim)
-            .collect { case Description(distance) => distance }
-            .foreach { distance => version += s"-$distance" }
+        V(version) match {
+          case None =>
+            sys.error(s"verion $version not formatted properly.")
+          case Some(value) =>
+            if (!(value.isSameSeries(baseV) || value >= baseV))
+              sys.error(s"Your latest tag $version cannot be less than tlBaseVersion $baseV")
```
armanbilge, line 93:
```suggestion
              sys.error(s"Your current version $version cannot be less than tlBaseVersion $baseV")
```

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
```diff
@@ -82,17 +82,25 @@ object TypelevelVersioningPlugin extends AutoPlugin {
...
         var version = latestInSeries.fold(tlBaseVersion.value)(_.toString)
 
-        // Looks for the distance to latest release in this series
-        latestInSeries.foreach { latestInSeries =>
-          Try(s"git describe --tags --match v$latestInSeries".!!.trim)
-            .collect { case Description(distance) => distance }
-            .foreach { distance => version += s"-$distance" }
+        V(version) match {
+          case None =>
+            sys.error(s"verion $version not formatted properly.")
```
armanbilge, line 90:
```suggestion
            sys.error(s"version must be semver format: $version")
```

armanbilge (commented):
Thanks, this is looking really good! Just some nits :)

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
osleonard: we could however the if would in-exhuastive is that ok ?

armanbilge: Yes, when the `if` doesn't match it just means the code should continue :) essentially we are writing

```scala
// do some stuff
if (isInvalid(stuff)) throw new Exception
// do more stuff
if (isInvalid(moreStuff)) throw new Exception
// return final answer
```

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
```diff
@@ -82,17 +82,26 @@ object TypelevelVersioningPlugin extends AutoPlugin {
...
         var version = latestInSeries.fold(tlBaseVersion.value)(_.toString)
 
+        V(version) match {
+          case None =>
+            sys.error(s"version must be semver format: $version")
+          case Some(value) =>
+            if (!(value.isSameSeries(baseV) || value >= baseV))
+              sys.error(
+                s"Your current version $version cannot be less than tlBaseVersion $baseV")
+        }
```
armanbilge, line 95: I totally missed this: actually, we need to move this check outside of the current `{ }`. Otherwise it's missing the current tags which are read here:
https://github.com/typelevel/sbt-typelevel/blob/b3b9fc2e6aac476260836bc91462468fb9c1a5aa/versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala#L66

armanbilge:
Thanks for working on this! :)

Issue #147, opened 2022-02-08
`base64: invalid input` due to newlines in `PGP_SECRET`


armanbilge:
This came up in https://github.com/erikerlandson/coulomb/issues/235. I've actually encountered it myself before and I think the behavior is different across OSes.

H/t to @erikerlandson for proposing two good solutions:

> 1. update sbt-typelevel quickstart to say gpg --armor --export-secret-keys $LONG_ID | base64 -w0 (i.e. tell base64 to not split the output into lines)
or
> 2. change the github action to use base64 -di, to ignore the newlines

In https://github.com/erikerlandson/coulomb/issues/235 I suggested we could apply both, but I'm not sure what the cross-OS support is for the `-w0` flag for `base64` suggested in (1).

However, since publish jobs are (effectively) always run from ubuntu, solution (2) should be robust.

armanbilge:
Yeah, macOS doesn't support the `-w` flag.

https://stackoverflow.com/questions/46463027/base64-doesnt-have-w-option-in-mac

So solution (2) would be better.

> However, since publish jobs are (effectively) always run from ubuntu

I think this is a reasonable requirement of sbt-typelevel. This doesn't preclude cross-OS builds or OS-specific artifacts. All it means is that the final job which downloads the intermediate artifacts and uploads them to sonatype will run on ubuntu.

erikerlandson:
it seems reasonable to assume/require that `publish` runs on some flavor of linux container :+1:

armanbilge:
Unfortunately the `-i` flag means something different on macOS 😕 

> However, since publish jobs are (effectively) always run from ubuntu, solution (2) should be robust.

> it seems reasonable to assume/require that `publish` runs on some flavor of linux container 👍

Finally someone needed to publish from macOS 😅 

- https://github.com/typelevel/sbt-typelevel/pull/648

PR #149, opened 2022-02-09
Ignore garbage when decoding PGP secret

armanbilge:
Fixes https://github.com/typelevel/sbt-typelevel/issues/147.

```
  -i, --ignore-garbage  when decoding, ignore non-alphabet characters
```


armanbilge:
For better or for worse the real test is when we merge this and it tries to release.

Issue #140, opened 2022-02-03
`Skipping update requested, but update has not previously run successfully.`


armanbilge:
Well, of course https://github.com/typelevel/sbt-typelevel/pull/117 has come back to bite us 😕 

I updated some projects to the latest hash and started getting this error. Search leads me to https://github.com/sbt/sbt/issues/5056 which suggests that our `skip` trickery is the culprit.

armanbilge:
Ok, so far this has affected two projects 😩

PR #162, opened 2022-02-10
Don't `skip` the `update` task

armanbilge:
Fixes https://github.com/typelevel/sbt-typelevel/issues/140. (hopefully :grimacing:)

Unfortunately `update / skip` is (almost always) a fatal error, and according to https://github.com/sbt/sbt/issues/5056 that is unlikely to change. So, we don't skip it.


armanbilge:
This fixes the update issue, but fs2-grpc then gets tripped up on the code-generating task. I think we have to revert https://github.com/typelevel/sbt-typelevel/pull/117.

PR #163, opened 2022-02-11
More responsible skipping

### LINKED ISSUE #140: `Skipping update requested, but update has not previously run successfully.`

armanbilge:
Fixes #140, this time for real.

Two changes:
1. Don't set `skip := true` for a whole project. Instead set it task-by-task, for tasks where skipping is known to be "safe".
2. Make this feature opt-in. More builds broke than builds that actually need this. It's a good default in theory, but in practice there are too many complex interactions that makes stuff break in confusing ways.

### DIFF
--- a/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
+++ b/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
@@ -48,15 +48,22 @@ object TypelevelKernelPlugin extends AutoPlugin {
       addCommandAlias("tlReleaseLocal", mkCommand(List("reload", "project /", "+publishLocal")))
 
   override def projectSettings = Seq(
-    skip := {
-      skip.value || {
+    skipIfIrrelevant(compile),
+    skipIfIrrelevant(test),
+    skipIfIrrelevant(publishLocal),
+    skipIfIrrelevant(publish)
+  )
+
+  private[sbt] def mkCommand(commands: List[String]): String = commands.mkString("; ", "; ", "")
+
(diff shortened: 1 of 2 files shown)


kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
```diff
@@ -48,15 +48,27 @@ object TypelevelKernelPlugin extends AutoPlugin {
...
+    },
+    skipIfIrrelevant(compile),
+    skipIfIrrelevant(test),
+    skipIfIrrelevant(publishLocal),
+    skipIfIrrelevant(publish)
+  )
+
+  private[sbt] def mkCommand(commands: List[String]): String = commands.mkString("; ", "; ", "")
+
+  private[sbt] def skipIfIrrelevant[T](task: TaskKey[T]) = {
```
armanbilge, line 64: I kept this private, but IMO it's worth exposing. For example http4s could use this for the `unusedCompileDependenciesTest` that was causing problems for the non-Scala 3 modules.

armanbilge: Although, many tasks are not natively `skip`pable (like `test`).

rossabaker: Yeah, seems worth exposing.

rossabaker (approved):

Issue #107, opened 2022-01-28
typelevel/download-java action for JVMs


armanbilge:
Currently, sbt-gh-actions works with JVMs supported natively by [actions/setup-java](https://github.com/actions/setup-java) plus GraalVM via [DeLaGuardo/setup-graalvm](https://github.com/DeLaGuardo/setup-graalvm).

I'd like to propose switching to [sdkman/sdkman-action](https://github.com/sdkman/sdkman-action) for GraalVM and other "alternate" JVMs.
1. It supports an impressive amount of JVMs, see output of `sdk list java` below.
2. It directly integrates with actions/setup-java by spitting out a `jdkFile` that is consumed by setup-java, and thus all the actual installation/setup is done by setup-java.
3. The bus factor is almost definitely higher, and SDKMAN is "recognized" by Scala itself: https://github.com/lampepfl/dotty/pull/13885

Pinging @vasilmkd for his thoughts.

```
================================================================================
Available Java Versions for Linux 64bit
================================================================================
 Vendor        | Use | Version      | Dist    | Status     | Identifier
--------------------------------------------------------------------------------
 Corretto      |     | 17.0.2.8.1   | amzn    |            | 17.0.2.8.1-amzn     
               |     | 17.0.1.12.1  | amzn    |            | 17.0.1.12.1-amzn    
               |     | 17.0.0.35.1  | amzn    |            | 17.0.0.35.1-amzn    
               |     | 11.0.14.9.1  | amzn    |            | 11.0.14.9.1-amzn    
               |     | 11.0.13.8.1  | amzn    |            | 11.0.13.8.1-amzn    
               |     | 11.0.12.7.1  | amzn    |            | 11.0.12.7.1-amzn    
               |     | 8.322.06.2   | amzn    |            | 8.322.06.2-amzn     
               |     | 8.312.07.1   | amzn    |            | 8.312.07.1-amzn     
 Dragonwell    |     | 11.0.12.8    | albba   |            | 11.0.12.8-albba     
               |     | 8.8.9        | albba   |            | 8.8.9-albba         
 GraalVM       |     | 22.0.0.2.r17 | grl     |            | 22.0.0.2.r17-grl    
               |     | 22.0.0.2.r11 | grl     |            | 22.0.0.2.r11-grl    
               |     | 21.3.1.r17   | grl     |            | 21.3.1.r17-grl      
               |     | 21.3.1.r11   | grl     |            | 21.3.1.r11-grl      
[...]

PR #168, opened 2022-02-18
Use typelevel jdk index for managing Java

armanbilge:
Closes https://github.com/typelevel/sbt-typelevel/issues/107. Integrates sbt-typelevel with [typelevel/jdk-index](https://github.com/typelevel/jdk-index) and [typelevel/download-java](https://github.com/typelevel/download-java).

Not all the distributions currently made available in the plugin are included in the typelevel/jdk-index, so these fallback to the old workflow steps. And similar for specific versions of GraalVM, since the index always points to the latest.

There will be minor breakage if someone was specifying a _full_ Java version (`"17"` vs `"17.0.2"`) but that's probably pretty unusual and not a good idea in general. We may want to consider requiring an `Int` here in the future instead of a `String`.

cc @vasilmkd

### DIFF
--- a/github-actions/src/main/scala/org/typelevel/sbt/gha/JavaSpec.scala
+++ b/github-actions/src/main/scala/org/typelevel/sbt/gha/JavaSpec.scala
@@ -26,7 +26,10 @@ final case class JavaSpec(dist: JavaSpec.Distribution, version: String) {
 object JavaSpec {
 
   def temurin(version: String): JavaSpec = JavaSpec(Distribution.Temurin, version)
+  def corretto(version: String): JavaSpec = JavaSpec(Distribution.Corretto, version)
   def graalvm(version: String): JavaSpec = JavaSpec(Distribution.GraalVM, version)
+  def openj9(version: String): JavaSpec = JavaSpec(Distribution.OpenJ9, version)
+  def oracle(version: String): JavaSpec = JavaSpec(Distribution.Oracle, version)
 
   @deprecated(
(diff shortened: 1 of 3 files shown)


armanbilge:
Also almost definitely not a big deal, but maybe time to admit that this should go to 0.5.0 🙃 

```
[error] sbt-typelevel-github-actions: Failed binary compatibility check against org.typelevel:sbt-typelevel-github-actions:0.4.3 (e:info.versionScheme=early-semver, e:sbtVersion=1.0, e:scalaVersion=2.12)! Found 1 potential problems
[error]  * the type hierarchy of object org.typelevel.sbt.gha.JavaSpec#Distribution#GraalVM is different in current version. Missing types {scala.runtime.AbstractFunction1}
[error]    filter with: ProblemFilters.exclude[MissingTypesProblem]("org.typelevel.sbt.gha.JavaSpec$Distribution$GraalVM$")
```

github-actions/src/main/scala/org/typelevel/sbt/gha/JavaSpec.scala
```diff
@@ -50,7 +50,7 @@ object JavaSpec {
   object Distribution {
     case object Temurin extends TlDistribution("temurin")
     case object Corretto extends TlDistribution("corretto")
-    case object GraalVM extends TlDistribution("graalvm")
+    case object LatestGraalVM extends TlDistribution("graalvm")
```
armanbilge, line 53: Eh. Let's us get it into 0.4.6.

Issue #79, opened 2022-01-18
Add JDK target setting


armanbilge:
This one is inspired by @ahjohannessen in https://github.com/armanbilge/fs2-grpc/pull/1#discussion_r786559343.

It would default to JDK 8, of course :) We could just hardcode this, but I think it would be nice if our settings plugin is friendly to applications that know they will be running on a newer JVM.

armanbilge:
As always, Vasil is doing some magic in https://github.com/typelevel/cats-effect/pull/2825 :)

What's interesting there is the use of the `--release` flag instead of the usual `-source`/`-target`.

This seems good for the next release, since it pairs nicely with https://github.com/typelevel/sbt-typelevel/issues/138.

PR #170, opened 2022-02-18
Add support for the JVM release option for scalac and javac

vasilmkd:


armanbilge:
I think this closes https://github.com/typelevel/sbt-typelevel/issues/79 :)

settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
```diff
@@ -29,13 +29,17 @@ object TypelevelSettingsPlugin extends AutoPlugin {
   object autoImport {
     lazy val tlFatalWarnings =
       settingKey[Boolean]("Convert compiler warnings into errors (default: false)")
+    lazy val tlJvmRelease =
+      settingKey[Int](
```
armanbilge, line 33: There's no universe where this somehow can no longer be represented as an `Int` right? 😅

settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
```diff
@@ -201,7 +205,29 @@ object TypelevelSettingsPlugin extends AutoPlugin {
         Seq("-Werror")
       else
         Seq.empty
+    },
+    scalacOptions ++= {
+      val releaseOption = if (isJava8) Seq() else Seq("-release", tlJvmRelease.value.toString)
+      val targetOption = if (isJava8) Seq() else Seq(s"-target:${tlJvmRelease.value}")
+
+      scalaVersion.value match {
```
armanbilge, line 213: I know it's ancient but Scala 2.11 supports this too.
https://github.com/typelevel/cats-parse/blob/7b524daf2abe63ca896656cfd36c13e9051476f6/build.sbt#L87

armanbilge: (I wouldn't ask if there weren't some TL projects still on 2.11).

settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
```diff
@@ -201,7 +205,29 @@ object TypelevelSettingsPlugin extends AutoPlugin {
         Seq("-Werror")
       else
         Seq.empty
+    },
+    scalacOptions ++= {
+      val releaseOption = if (isJava8) Seq() else Seq("-release", tlJvmRelease.value.toString)
+      val targetOption = if (isJava8) Seq() else Seq(s"-target:${tlJvmRelease.value}")
```
armanbilge, line 211: On 2.12.15 it seems the formatting is different?
```
  -target:<target>                     Target platform for object files. All JVM 1.5 - 1.7 targets are deprecated. (jvm-1.5,jvm-1.6,jvm-1.7,[jvm-1.8])
```

settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
vasilmkd: Done. It's a bit opinionated as nothing before JDK 8 is supported by `sbt-typelevel`. I hope that's ok.

settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
```diff
@@ -29,13 +29,17 @@ object TypelevelSettingsPlugin extends AutoPlugin {
   object autoImport {
     lazy val tlFatalWarnings =
       settingKey[Boolean]("Convert compiler warnings into errors (default: false)")
+    lazy val tlJvmRelease =
```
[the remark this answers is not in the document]
armanbilge: Oh, interesting, javac uses JVM.
```
  -target <release>            Generate class files for specific VM version
```

settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
armanbilge: Yeah, I think that's more than reasonable.

settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
vasilmkd: Done.

armanbilge (dismissed):
Yay! Thank you :)

armanbilge:
I've been thinking about how we can get this PR in to the 0.4.x series. Since CI is running on JDK 8 anyway, I guess the point of this PR is to support projects that are published manually, from possibly newer JDKs?

If that's the case, I propose we actually _don't_ set a default value for this setting. If this setting is not set, then no flags are added; but if it is set, it works exactly like you've got it here. So for the projects release manually this can be set to 8 as needed.

We can create a follow-up issue to default it to 8 when we break compat in the next minor bump.

WDYT?

vasilmkd:
Nothing that's added here works for fs2 and essentially breaks it. We should put this on ice.

armanbilge:
Right, by ice you mean until the next breaking release, yes?

Fast-forwarding to then, I'm still concerned whether this works for fs2 at all.

vasilmkd:
It doesn't, if this ships, fs2 would have to disable all of it.

armanbilge:
So then maybe the setting should be an `Option[Int]`? Where `None` disables it?

settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
```diff
@@ -29,13 +29,17 @@ object TypelevelSettingsPlugin extends AutoPlugin {
...
+      settingKey[Option[Int]](
+        "JVM target version for the compiled bytecode, None results in default scalac and javac behavior (no --release flag is specified). (default: Some(8), supported values: 8, 9, 10, 11, 12, 13, 14, 15, 16, 17)")
   }
 
   import autoImport._
   import TypelevelKernelPlugin.autoImport._
 
   override def globalSettings = Seq(
     tlFatalWarnings := false,
+    tlJdkRelease := Some(8),
```
armanbilge, line 42: Not to be too annoying, but we can make this `None` and get this into `series/0.4` :)

vasilmkd:
Yeah, sorry, I got distracted.

settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
```diff
@@ -200,7 +204,42 @@ object TypelevelSettingsPlugin extends AutoPlugin {
         Seq("-Werror")
       else
         Seq.empty
+    },
+    scalacOptions ++= {
+      val releaseOption = tlJdkRelease
+        .value
+        .map { release => if (isJava8) Seq.empty else Seq("-release", release.toString) }
```
armanbilge, line 211: Why do we need the `isJava8` check for `scalacOptions`?

Also, what is the failure mode if `tlJdkRelease > 8` but `java.version == 1.8`?

vasilmkd: If `java.version == 1.8` it means we're running `javac 1.8` which doesn't understand the `--release` flag. `scalac` somehow makes use of this and also fails compilation with
```
[error] '8' is not a valid choice for '-target'
[21](https://github.com/typelevel/cats-effect/runs/5215692131?check_suite_focus=true#step:12:21)
[error] '8' is not a valid choice for '-target'
[22](https://github.com/typelevel/cats-effect/runs/5215692131?check_suite_focus=true#step:12:22)
[error] bad option: '-target:8'
[23](https://github.com/typelevel/cats-effect/runs/5215692131?check_suite_focus=true#step:12:23)
[error] bad option: '-target:8'
[24](https://github.com/typelevel/cats-effect/runs/5215692131?check_suite_focus=true#step:12:24)
[error] (kernelJVM / Compile / compileIncremental) Compilation failed
[25](https://github.com/typelevel/cats-effect/runs/5215692131?check_suite_focus=true#step:12:25)
[error] (kernelJS / Compile / compileIncremental) Compilation failed
[26](https://github.com/typelevel/cats-effect/runs/5215692131?check_suite_focus=true#step:12:26)
[error] Total time: 13 s, completed Feb 16, 2022 11:57:41 AM
[27](https://github.com/typelevel/cats-effect/runs/5215692131?check_suite_focus=true#step:12:27)

[28](https://github.com/typelevel/cats-effect/runs/5215692131?check_suite_focus=true#step:12:28)
Error: Process completed with exit code 1.
```

from https://github.com/typelevel/cats-effect/runs/5215692131?check_suite_focus=true

armanbilge: Oh, cool, that seems good. But if we don't add the flag, it won't protected us from a target JDK newer the the compile-time JDK since it basically ignores that setting, so it seems we need another check?

vasilmkd: JDK 8 cannot target a newer JDK. Am I understanding your question correctly?

vasilmkd: If you don't specify anything, `javac 1.8` compiles JVM 8 bytecode, 17 compiles JVM 17 bytecode, while scalac compiles JVM 8 bytecode.

armanbilge: Apologies if I'm not reading your code correctly. Here's my hypothetical situation:
1. `tlJdkRelease := Some(17)`
2. I load my project in sbt on Java 8
3. No release flags are added to `scalacOptions`, b/c `java.version == 1.8`
4. My project compiles, even though I asked to target JDK 17.

vasilmkd: If a project is specifically targetting 17 and the code is compilable by JDK 8, I don't know why they're targetting 17 in the first place. 😆

armanbilge: That's a fair point :)

`scalac` can generate more optimized bytecode in certain situations when targeting newer JDKs.

Also, the fact that your project would fail to compile if you load it in Java 9 (but 8 would work!) is a bit odd imo :)

armanbilge: e.g.
- https://github.com/lampepfl/dotty/pull/12929

settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
```diff
@@ -209,15 +209,27 @@ object TypelevelSettingsPlugin extends AutoPlugin {
     scalacOptions ++= {
       val releaseOption = tlJdkRelease
         .value
-        .map { release => if (isJava8) Seq.empty else Seq("-release", release.toString) }
+        .map {
+          case 8 if isJava8 => Seq.empty
+          case n if n >= 8 => Seq("-release", n.toString)
+          case n => sys.error(s"'$n' is not a valid choice for '-release'")
+        }
```
armanbilge, line 216: Sorry to nitpick :(

This doesn't seem very DRY :) also I wonder if an error message specifically referencing the `tlJdkRelease` setting might do more to point users in the right direction. Also, there's a difference between "your JDK is too old" vs "n < 8 is not a valid choice".

Thanks again for your work on this ❤️

vasilmkd: Done.

settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
```diff
@@ -201,7 +205,56 @@ object TypelevelSettingsPlugin extends AutoPlugin {
...
+      case n if n >= 8 =>
+        if (javaRuntimeVersion < n) {
+          sys.error(
+            s"Target JDK is $n but you are using an older JDK $javaRuntimeVersion. Please switch to JDK >= $n.")
+        } else {
+          f(n)
+        }
+      case n =>
+        sys.error(
+          s"You're using JDK $n, which is not supported by `sbt-typelevel`. Please switch to a newer JDK.")
```
armanbilge, line 250: `n` is the target JDK right, not the user's runtime JDK?

```suggestion
          s"Target JDK is $n, which is not supported by `sbt-typelevel`. Please select a JDK >= 8.")
```

armanbilge:
@vasilmkd looks like they are discussing renaming these flags in Scala 3.
https://contributors.scala-lang.org/t/improving-scala-3-forward-compatibility/5298/22

PR #172, opened 2022-02-19
Optimize sorting in the `GitHelper.previousReleases`

danicheg:
Let's save a few nanoseconds on the loading of the project. Though it's only the hypothesis that sorting using git is faster. But I can bet it's so.

### DIFF
--- a/kernel/src/main/scala/org/typelevel/sbt/kernel/GitHelper.scala
+++ b/kernel/src/main/scala/org/typelevel/sbt/kernel/GitHelper.scala
@@ -27,15 +27,14 @@ private[sbt] object GitHelper {
    * @param fromHead
    *   if `true`, only tags reachable from HEAD's history. If `false`, all tags in the repo.
    */
-  def previousReleases(fromHead: Boolean = false): List[V] = {
+  def previousReleases(fromHead: Boolean = false): List[V] =
     Try {
       val merged = if (fromHead) " --merged HEAD" else ""
       // --no-contains omits tags on HEAD
-      s"git tag --no-contains HEAD$merged".!!.split("\n").toList.map(_.trim).collect {
+      s"git tag --no-contains HEAD$merged --sort=-v:refname # reverse".!!.split("\n").toList.map(_.trim).collect {
         case V.Tag(version) => version
       }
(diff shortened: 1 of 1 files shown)


kernel/src/main/scala/org/typelevel/sbt/kernel/GitHelper.scala
```diff
@@ -27,15 +27,17 @@ private[sbt] object GitHelper {
...
+  def previousReleases(fromHead: Boolean = false): List[V] =
     Try {
       val merged = if (fromHead) " --merged HEAD" else ""
       // --no-contains omits tags on HEAD
-      s"git tag --no-contains HEAD$merged".!!.split("\n").toList.map(_.trim).collect {
-        case V.Tag(version) => version
-      }
-    }.getOrElse(List.empty).sorted.reverse
-  }
+      s"git tag --no-contains HEAD$merged --sort=-v:refname # reverse"
```
armanbilge, line 34: It's probably fine but makes me nervous! 😆 
```suggestion
      s"git tag --no-contains HEAD$merged --sort=-v:refname" // reverse
```

armanbilge:
Hmm. I ran this in terminal and I'm not sure it's the right order. Shouldn't the v0.4.0 milestones come before it?
```
$ git tag --no-contains HEAD --sort=-v:refname
v0.4.5
v0.4.4
v0.4.3
v0.4.2
v0.4.1
v0.4.0-M5
v0.4.0-M4
v0.4.0-M3
v0.4.0-M2
v0.4.0-M1
v0.4.0
```

danicheg:
Well, then we should try something like
```
git -c 'versionsort.suffix=-' tag --no-contains HEAD --sort=-v:refname
```
output is
```
v0.4.3
v0.4.2
v0.4.1
v0.4.0
v0.4.0-M5
v0.4.0-M4
v0.4.0-M3
v0.4.0-M2
v0.4.0-M1
v0.3.2
v0.3.1
v0.3
v0.2
v0.1
```

armanbilge:
That looks good!

armanbilge (approved):
Very nice, thank you! This is also a partial fix for https://github.com/typelevel/sbt-typelevel/issues/14.

PR #173, opened 2022-02-20
Micro-optimizations of using collections

danicheg:
Some optimizations to silence the $IDE linter mostly.

### DIFF
--- a/versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
+++ b/versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
@@ -74,10 +74,12 @@ object TypelevelVersioningPlugin extends AutoPlugin {
           .previousReleases(true)
           .filterNot(_.isPrerelease) // TODO Ordering of pre-releases is arbitrary
           .headOption
-          .flatMap { previous =>
-            if (previous > baseV)
+          .flatMap {
+            case previous if previous > baseV =>
               sys.error(s"Your tlBaseVersion $baseV is behind the latest tag $previous")
-            else Some(previous).filter(baseV.isSameSeries)
+            case previous if baseV.isSameSeries(previous) =>
+              Some(previous)
+            case _ => None
           }
 
         // version here is the prefix used further to build a final version number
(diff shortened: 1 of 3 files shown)


versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
```diff
@@ -77,10 +77,7 @@ object TypelevelVersioningPlugin extends AutoPlugin {
           .flatMap { previous =>
             if (previous > baseV)
               sys.error(s"Your tlBaseVersion $baseV is behind the latest tag $previous")
-            else if (baseV.isSameSeries(previous))
-              Some(previous)
-            else
-              None
+            else Some(previous).filter(baseV.isSameSeries)
```
armanbilge, line 80: I wonder if this one is more confusing now.

armanbilge (approved):
One thought, otherwise nice cleanups 👍 thank you! Or thank your linter for me 😉

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
danicheg: I could advise another variant:
```scala
.flatMap { 
  case previous if previous > baseV =>
     sys.error(s"Your tlBaseVersion $baseV is behind the latest tag $previous")
  case previous =>
    Option(previous).filter(baseV.isSameSeries)
}
```
I often use this `Option(foo).filter(_ => ???)` pattern. But if it confuses you, then let's undo it. I'm just not so much like nested ifs tbh.

armanbilge: Yeah for some reason `Option(foo).filter(_ => ???)` confuses me here 😅 there are 3 cases, and it sort of merges 2 of them into one case.

I would be okay with this. It's basically the same as the if-else.
```scala
case previous if previous > baseV =>
   sys.error(s"Your tlBaseVersion $baseV is behind the latest tag $previous")
case previous if baseV.isSameSeries(previous) =>
  Some(previous)
case _ => None 
```

danicheg: Roger that 👍🏻

armanbilge:
A bona-fide regular contributor! 🎉

Issue #174, opened 2022-02-20
`error: invalid key: 'versionsort.suffix`


armanbilge:
Seems to be due to https://github.com/typelevel/sbt-typelevel/pull/172, cc @danicheg.

```
[info] loading settings for project sbt-typelevel-build from build.sbt,ci-signing.sbt,ci.sbt,core.sbt,github-actions.sbt,github.sbt,mima.sbt,settings.sbt,site.sbt,sonatype.sbt,versioning.sbt ...
[info] loading project definition from /workspace/sbt-typelevel/project
[info] loading settings for project root from build.sbt ...
[info] loading settings for project settings from build.sbt ...
[info] loading settings for project github from build.sbt ...
[info] loading settings for project githubActions from build.sbt ...
[info] loading settings for project versioning from build.sbt ...
[info] loading settings for project mima from build.sbt ...
[info] loading settings for project sonatype from build.sbt ...
[info] loading settings for project ciSigning from build.sbt ...
[info] loading settings for project ci from build.sbt ...
[info] loading settings for project core from build.sbt ...
[info] loading settings for project site from build.sbt ...
[info] resolving key references (17613 settings) ...
error: invalid key: 'versionsort.suffix
fatal: unable to parse command-line config
error: invalid key: 'versionsort.suffix
fatal: unable to parse command-line config
error: invalid key: 'versionsort.suffix
fatal: unable to parse command-line config
error: invalid key: 'versionsort.suffix
fatal: unable to parse command-line config
error: invalid key: 'versionsort.suffix
fatal: unable to parse command-line config
error: invalid key: 'versionsort.suffix
fatal: unable to parse command-line config
error: invalid key: 'versionsort.suffix
fatal: unable to parse command-line config
error: invalid key: 'versionsort.suffix
fatal: unable to parse command-line config
error: invalid key: 'versionsort.suffix
fatal: unable to parse command-line config
error: invalid key: 'versionsort.suffix
fatal: unable to parse command-line config
error: invalid key: 'versionsort.suffix
fatal: unable to parse command-line config
error: invalid key: 'versionsort.suffix
fatal: unable to parse command-line config
error: invalid key: 'versionsort.suffix
fatal: unable to parse command-line config
error: invalid key: 'versionsort.suffix
fatal: unable to parse command-line config
error: invalid key: 'versionsort.suffix
fatal: unable to parse command-line config
error: invalid key: 'versionsort.suffix
fatal: unable to parse command-line config
[...]

PR #175, opened 2022-02-20
Fix for #174

danicheg:


kernel/src/main/scala/org/typelevel/sbt/kernel/GitHelper.scala
```diff
@@ -31,7 +31,7 @@ private[sbt] object GitHelper {
     Try {
       val merged = if (fromHead) " --merged HEAD" else ""
       // --no-contains omits tags on HEAD
-      s"git -c 'versionsort.suffix=-' tag --no-contains HEAD$merged --sort=-v:refname" // reverse
+      s"git -c versionsort.suffix=- tag --no-contains HEAD$merged --sort=-v:refname" // reverse
```
armanbilge, line 34: 😵

armanbilge (approved):
Confirmed it fixes it on my machine. Thanks for the fast response!

PR #176, opened 2022-02-20
Add `tlMimaPreviousVersions` setting

armanbilge:
We previously debated its usefulness in https://github.com/typelevel/sbt-typelevel/pull/46, but now I'm convinced.

1. `mimaPreviousArtifacts` is at the project level, but often you want to filter botched releases at the build-level. `tlMimaPreviousVersions` is build-level.
2. Some projects add extra versions, like cats includes the 1.x series for some modules.
3. Some projects are just weird, like scalacheck has this pvp version thing going on.

So yeah, I'm convinced :)

### DIFF
--- a/mima/src/main/scala/org/typelevel/sbt/TypelevelMimaPlugin.scala
+++ b/mima/src/main/scala/org/typelevel/sbt/TypelevelMimaPlugin.scala
@@ -47,27 +47,23 @@ object TypelevelMimaPlugin extends AutoPlugin {
         versionScheme.value.contains("early-semver"),
         "Only early-semver versioning scheme supported.")
 
-      if (publishArtifact.value) {
-        val current = V(version.value)
-          // Consider it as a real release, for purposes of compat-checking
-          .map(_.copy(prerelease = None))
-          .getOrElse(sys.error(s"Version must be semver format: ${version.value}"))
+      val current = V(version.value)
+        // Consider it as a real release, for purposes of compat-checking
+        .map(_.copy(prerelease = None))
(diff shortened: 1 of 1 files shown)


danicheg (commented):
Sorry for being late on this. Here are my 2 cents on the topic. Wouldn't that be confusing for users to deal with the settings behaving likewise the upstream settings? Say, the `tlFoo` setting where `foo` comes from an upstream library. An ability to configure things at the build/project level multiple ways increases entropy, I guess. Moreover, if new `tl-*` settings would slightly differ in behavior against original upstream settings, then it really can be hard to figure out things.

armanbilge:
@danicheg thanks for your feedback, always much much appreciated!

I guess you are talking about the difference between `mimaPreviousArtifacts` and `tlMimaPreviousVersions`?

The problem is that `mimaPreviousArtifacts` is very difficult to configure correctly (remember that mouse had misconfigured this, so JS artifacts were being compared to JVM artifacts?). In fact many projects misconfigured this which is one of the reason I was motivated to make sbt-typelevel.

I agree that it is confusing, and I would like to make this better though. WDYT?

danicheg:
Nevermind. I'm just overthinking about corner cases, probably. sbt-typelevel can mitigate some difficulties of configuring plugins for their users. As you said, users might be wrong on it also without sbt-typelevel.

armanbilge:
FWIW I expect this setting to be rarely used. Of all the Typelevel projects, really only Cats and Scalacheck seem to need it.

armanbilge:
Darn, this PR introduced a regression.

Issue #144, opened 2022-02-06
Create a Typelevel Laika Theme


armanbilge:
As very nicely documented in https://github.com/planet42/Laika/discussions/253#discussioncomment-2119387.

It seems that this is a much more principled way to provide default settings/config for Laika.

PR #180, opened 2022-02-20
Create a Typelevel helium theme

armanbilge:
Closes https://github.com/typelevel/sbt-typelevel/issues/144. Fixes https://github.com/typelevel/sbt-typelevel/issues/121.

@jenshalm if you happen to have a chance, I would be very grateful for your eyes on this 🙏 Thank you again for the guide in https://github.com/planet42/Laika/discussions/253, it was very very helpful ❤️ 

@danicheg since you are passionate about Typelevel microsites, I'd love to hear your ideas as well!

Screen Shot 2022-02-20 at 13 42 55


site/src/main/resources/org/typelevel/sbt/helium/default.template.html
```diff
@@ -0,0 +1,107 @@
...
+        ]
+      }
+      
+      <ul class="nav-list">
+        <li class="level1 nav-header">Related Projects</li>
+        <li class="level2"><a href="https://typelevel.org/cats/">cats</a></li>
+        <li class="level2"><a href="https://typelevel.org/cats-effect/">cats-effect</a></li>
+        <li class="level2"><a href="https://fs2.io/">fs2</a></li>
+        <li class="level2"><a href="https://scalacheck.org/">scalacheck</a></li>
+      </ul>
```
armanbilge, line 75: @danicheg any thoughts on this one?

site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
```diff
@@ -84,6 +98,13 @@ object TypelevelSitePlugin extends AutoPlugin {
       Helium
```
armanbilge, line 98: @jenshalm I couldn't figure it if/how it's possible to move this config into the `TypelevelHeliumExtensions`, would appreciate any pointers.

site/src/main/resources/org/typelevel/sbt/helium/default.template.html
danicheg: Should be that bar be configurable? It's easy to imagine that a particular project could be not related to scalacheck/fs2/etc.

armanbilge: Yes, for sure, I was thinking about this too. I started by grabbing the "pinned" projects from the Typelevel org page.

We can make it configurable, or change it from "related" to something else? Or just list every project there? 😆 

If we make it configurable, then maybe we need add some helper method so its easy to add TL libraries there without checking the name, url etc.

armanbilge: I was thinking we can use a `Seq[(String, URL)` and provide something like this:
https://github.com/sbt/librarymanagement/blob/67988135152f88afa9134fcfc357c0901e700f40/core/src/main/scala/sbt/librarymanagement/License.scala#L10

danicheg: Sure. More type-safety is always preferable.

site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
```diff
@@ -16,24 +16,35 @@
...
+import gha.GenerativePlugin
+import GenerativePlugin.autoImport._
+
 object TypelevelSitePlugin extends AutoPlugin {
 
   object autoImport {
     lazy val tlSiteHeliumConfig = settingKey[Helium]("The Helium configuration")
     lazy val tlSiteApiUrl = settingKey[Option[URL]]("URL to the API docs")
+    lazy val tlSiteRelated =
+      settingKey[Seq[(String, URL)]]("A list of related projects (default: cats)")
```
armanbilge, line 46: @danicheg wdyt?

site/src/main/scala/org/typelevel/sbt/TypelevelProjects.scala
```diff
@@ -0,0 +1,33 @@
...
+
+object TypelevelProject {
+  val Cats = tl("cats")
+  val CatsEffect = tl("cats-effect")
+  val Discipline = gh("discipline")
+  val Fs2 = "fs2" -> url("https://fs2.io/")
+  val Http4s = "http4s" -> url("https://http4s.org/")
+  val Scalacheck = "scalacheck" -> url("https://scalacheck.org/")
+  val Shapeless = gh("shapeless", "milessabin")
+  val Shapeless3 = gh("shapeless-3")
```
armanbilge, line 29: I think probably these are the most important, but happy to add anything :)

site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
armanbilge: Maybe should be `tlSiteRelatedProjects` ?

site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
armanbilge: I took another look at this and I was able to do it :)

site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
```diff
@@ -71,7 +83,21 @@ object TypelevelSitePlugin extends AutoPlugin {
...
+        tlSiteApiUrl.value,
+        scmInfo.value.map(_.browseUrl),
+        licenses.value.headOption,
+        tlSiteRelated.value
+      )
+    },
+    tlSiteHeliumConfig := Helium.defaults,
+    laikaTheme := {
+      tlSiteHeliumConfig.value.build.extend(tlSiteTheme.value)
+    },
```
armanbilge, line 100: This configuration still feels awkward 😕

jenshalm: What would you prefer as an alternative here? 

I assume from the user's perspective they can completely ignore the `laikaTheme` setting here (which would need to be documented as they might be tempted to use it by looking at Laika docs) and the remaining 2 are the core Helium features and the Typelevel extra theme features which I feel are best kept separate. For one of the settings you'd add docs to your own site and for the other you could directly point them at the Laika page for theme settings.

The only way you could create a unified theme config API would be if the base theme would be a fairly small feature set that would enable you to re-expose its config API in your own theme API. But the Helium API is huge and replicating it in an extension is not feasible/desirable.

jenshalm: Now looking at the implementation I'd recommend not to completely hide the core Helium config inside the `TypelevelHeliumExtensions` class. The hidden defaults would override anything users would set in `tlSiteHeliumConfig`. You also kind of create the Helium theme twice (which would also lead to some things executing twice). I feel like it's more convenient and better separated if users could write something like this:

```scala
tlSiteHeliumBase = TypeLevelHeliumBase.defaults
  .withFoo
  .withBar...

tlSiteHeliumExtensions = TypeLevelHeliumExtensions.defaults
  .withFoo
  .withBar...
```

Where the first setting is the core Helium API as documented in the Laika manual, only that the defaults are already richer than the builtin `Helium.defaults` (essentially the `base` value currently hidden inside the extension impl) and the second setting is your own API with functionality not covered by core Helium.

This is certainly not ideal, but maybe the best option given that wrapping and re-exposing the extensive Helium API is not feasible?

armanbilge: > the hidden defaults would override anything users would set in `tlSiteHeliumConfig`. You also kind of create the Helium theme twice (which would also lead to some things executing twice).

Thanks, yes, those were exactly the problems I was concerned about. Actually this change was not backwards-compatible because of the overriding thing.

> This is certainly not ideal, but maybe the best option given that wrapping and re-exposing the extensive Helium API is not feasible?

You are right, I was trying to avoid having two separate settings for the base config and extensions. But it seems unavoidable, so I agree it is probably the best option.

> I assume from the user's perspective they can completely ignore the `laikaTheme` setting here (which would need to be documented as they might be tempted to use it

Right, well it depends whether the user is a Typelevel project or not. In fact, I don't expect most Typelevel projects to care about these settings at all. Only a few projects might need to make some very minor customization, except the sbt-http4s-org plugin which will add its own extensions.

https://github.com/typelevel/sbt-typelevel/blob/791f9a905ba0d54e800759a39f8609ecab62e6c7/build.sbt#L152-L156

https://github.com/http4s/http4s-dom/blob/cce807dfec6754e89b1eae73b306401b5b3997e6/build.sbt#L200-L204

 Non-typelevel projects can ignore these and make their own theme following the Laika manual.

jenshalm: Additional note: it might also be worthwhile to check whether the things you do in your extension settings are generic enough to be offered by the core Helium API, but that would be a thing for 0.19 and not something you can plan with for this PR.

PR #183, opened 2022-02-21
Add the `tlSitePublishTags` setting

### LINKED ISSUE #137: Proposal: explicit setting for disabling site publishing
This was born from a discussion https://github.com/typelevel/log4cats/pull/603#pullrequestreview-870754763.

The site plugin contains the `tlSitePublishBranch` setting which could have a `None` value. But that would mean the site will still publishing just only for tags. There are cases when building the site is fine, but publishing is redundant. It's the case of the multibranch project without a versioned site. A versioned site for such projects could be redundant at all. So it would be nice to have the ability to disable site publishing from plugin settings. 
On the other hand, one can set `tlSitePublishBranch` to the `main` branch from which the site will be publishing, but it would be more clear if one can explicitly disable the publishing site.
So this proposal is just about getting more configurable behavior through explicit settings rather than getting functionality at all.

danicheg:
Fixes #137. Implemented the way as @armanbilge suggested.

### DIFF
--- a/site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
+++ b/site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
@@ -42,6 +42,8 @@ object TypelevelSitePlugin extends AutoPlugin {
       "A sequence of workflow steps which publishes the site (default: peaceiris/actions-gh-pages)")
     lazy val tlSitePublishBranch = settingKey[Option[String]](
       "The branch to publish the site from on every push. Set this to None if you only want to update the site on tag releases. (default: main)")
+    lazy val tlSitePublishTags = settingKey[Boolean](
+      "Defines whether the site should be published on tag releases. Note on setting this to true requires the 'tlSitePublishBranch' setting to be set to None. (default: false)")
     lazy val tlSite = taskKey[Unit]("Generate the site (default: runs mdoc then laika)")
   }
 
@@ -53,6 +55,7 @@ object TypelevelSitePlugin extends AutoPlugin {
 
   override def buildSettings = Seq(
     tlSitePublishBranch := Some("main"),
+    tlSitePublishTags := tlSitePublishBranch.value.fold(true)(_ => false),
     tlSiteApiUrl := None,
     tlSiteKeepFiles := true,
     homepage := {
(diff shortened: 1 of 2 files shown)


site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
```diff
@@ -53,6 +55,7 @@ object TypelevelSitePlugin extends AutoPlugin {
 
   override def buildSettings = Seq(
     tlSitePublishBranch := Some("main"),
+    tlSitePublishTags := tlSitePublishBranch.value.fold(true)(_ => false),
```
armanbilge, line 58:
```suggestion
    tlSitePublishTags := tlSitePublishBranch.value.isEmpty
```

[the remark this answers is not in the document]
armanbilge: Yeah, thanks for preserving that behaviour. Probably in 0.5.0 we might want to make it `false` by default since it's a bit confusing.

site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
```diff
@@ -128,30 +131,50 @@ object TypelevelSitePlugin extends AutoPlugin {
...
+                GenerativePlugin.compileBranchPredicate("github.ref", predicate)
+              Some(s"github.event_name != 'pull_request' && $publicationCond")
+            }
+          )
+        )
+
+      val tlSitePublishTagsV = tlSitePublishTags.value
+      val tlSitePublishBranchV = tlSitePublishBranch.value
+
+      tlSitePublishTagsV -> tlSitePublishBranchV match {
```
armanbilge, line 162: Nit :) I only use `->` for key-value pairs.

```suggestion
      (tlSitePublishTagsV, tlSitePublishBranchV) match {
```

site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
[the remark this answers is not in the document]
danicheg: I 100% agree with this. Though behavior-changing is clumsy within a minor release, we are still not 1.0, so I see no evil.

armanbilge (approved):
Thanks for fixing this one, it was my fault to make it so confusing.

I think in general publishing site from tag can be a "thin ice" as you say. The problem is that tags can happen on _any_ branch, including old ones, or new development branch. So if you forget to disable this setting and don't setup versioned sites, then if you publish an early milestone tag you could accidentally override your site for the stable version.

site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
danicheg: So be it.

danicheg:
So, is this waiting for something?

armanbilge:
Because sbt-typelevel publishes the site on every push, and you updated the docs in this PR (thanks for that!), I'll merge this one right before I release so that the docs are not predicting the future 😉

armanbilge:
Thanks again! :)

Issue #186, opened 2022-02-23
Create dedicated documentation for typelevel-site-plugin


armanbilge:
With much thanks to @quafadas for guinea-pigging this, with a not-so-great experience.

Currently, the documentation consists of a brief FAQ and some bullet points on the customization page. In fact, this plugin probably deserves a page of its own.

The other important thing is that non-TL projects will want to configure the plugin differently than TL projects in order to override the default TL Laika theme. It seems the right thing to do is point them directly to the Laika docs and tell them which typelevel-specific settings to ignore.

It's also probably worth pointing out that this is an optional and opinionated plugin that bakes in Laika. Projects that don't want Laika and/or want to do very custom things should probably explore other microsite solutions instead of attempting to retrofit this plugin.

Quafadas:
I've been trying to think about what made this tricky, and is specific to this plugin, rather than my project. 

It is worth noting that I came in as a complete n00b... so some challenges were to be expected. Getting a n00b publishing to maven / own site as fast as I go there is impressive. So let's not forget that. 

Artefact publishing worked startlingly smoothly - it was a killer feature which kept me going through some docs troubles.

Publishing the doc website, and even Laika itself worked really well, _once I got it working_. 

I think that shows that the convenience this plugin targets, is actually not that far off. And is a super attractive goal. 

I did hit a couple of issues. Many may be idiosyncratic to my library which deliberately tries to cross the JS / JVM boundaries, which is in general a mess and bad idea :-), let's not write about there here. None-the-less, my suggestions: 

One
I followed the docs on the Laika website, but it didn't have the effects advertised. It was not obvious to me that tlXXX settings, suggested by Arman, and the Laika SBT plugin settings advertised in Laika docs, would trample each other. Some of the behaviour become seemingly non-deterministic ( I suspect because depending on the position of settings in build.sbt then changed it's behaviour), when changing things, this was difficult. 

It's not obvious to me (outside of reading the plugin code) how I could have found which settings interact and trample each other. Hence, for a non-TL user (who very likely wants to customise the colors / other settings), I would propose to try to find a way to make it possible to follow the Laika documentation for the SBT plugin directly, rather than use the TLHelium mechanisms. Ideally perhaps, somehow enforce "one paradigm or the other but not both" if that is possible.

Two
The FAQs suggesting running docs/laikaPreview
https://typelevel.org/sbt-typelevel/faq.html

[...]

PR #189, opened 2022-02-25
Add `tlSitePreview`

armanbilge:
This PR adds a single command `tlSitePreview` which runs both `mdoc --watch` and `laikaPreview` simultaneously, for a cohesive live-reload editing experience. Thanks to @Quafadas for reminding me about this in https://github.com/typelevel/sbt-typelevel/issues/186#issuecomment-1050321968.

cc @jenshalm (sorry for all the noise this week!): I thought you may be interested what this looks like, since you also mentioned this problem while working on the http4s site.

Unfortunately I had to copy the implementation of `laikaPreview` here. But, I wonder if the next version of Laika can define a `Task[() => ()]` `laikaStartPreview` that starts the server and returns the shutdown hook. `laikaPreview` can then be implemented using that, but it also makes it easier to integrate with other tooling like mdoc.


jenshalm:
Yes, I didn't think of that scenario. Can you create a ticket?

I'm also generally confused by the best way to create long-running tasks in sbt. When I had to evacuate to Windows for a MacBook repair I noticed I cannot use the same shortcut to stop the task (ctrl-D). I also often get a bunch of error messages from http4s when stopping the preview server and I'm not sure whether there is any way to prevent that.

PR #190, opened 2022-02-26
Add plugin for generating `.mergify.yml` config

armanbilge:
This was originally proposed in https://github.com/djspiewak/sbt-github-actions/issues/46.

Essentially, the problem is that in order to validate CI success the mergify config needs a list of _all_ the jobs in the CI matrix with their exact names (including the Scala version) ([mergify docs on this topic](https://docs.mergify.com/conditions/#validating-all-status-checks)). In practice, manually keeping these in sync is annoying, and sbt-gh-actions is managing the build matrix anyway.

By default, the plugin includes a configuration to merge CI-passing PRs from scala-steward for patch updates. This can be further configured to include minor bumps and/or disabled.

Another idea I have is to generate config for auto-labeling of PRs based on file paths of affected modules, like http4s does. It'll be nice not to have to hand-maintain that.

Since this plugin is new, I'm keeping it as a separate module that has to be added explicitly to a build. It's still lacking DSL for many mergify features, but we can add them as required. It currently supports merging and labeling which are probably the most important.

Once it's proven itself we can consider if we want to include it in the sbt-typelevel super-plugin.


.mergify.yml
```diff
@@ -0,0 +1,19 @@
...
+
+pull_request_rules:
+- name: merge scala-steward's PRs
+  conditions:
+  - author=scala-steward
+  - or:
+    - body~=labels:.*early-semver-patch
+    - body~=labels:.*early-semver-minor
+  - status-success=Build and Test (ubuntu-latest, 2.12.15, temurin@8, rootJVM)
+  - '#approved-reviews-by>=1'
```
armanbilge, line 16: @fthomas if you have a chance, would appreciate your eyes on this. Are these labels the right configuration for checking steward PRs? Thanks! :)

build.sbt
```diff
@@ -11,12 +11,16 @@ ThisBuild / developers := List(
   tlGitHubDev("djspiewak", "Daniel Spiewak")
 )
 
+ThisBuild / mergifyStewardConfig ~= { _.map(_.copy(mergeMinors = true)) }
+ThisBuild / mergifySuccessConditions += MergifyCondition.Custom("#approved-reviews-by>=1")
```
armanbilge, line 15: `mergeMinors`, because I'm dog-fooding. But I actually don't trust any update to upstream sbt plugins is not going to break things, patch or not ...

mergify/src/main/scala/org/typelevel/sbt/mergify/MergifyAction.scala
```diff
@@ -0,0 +1,65 @@
...
+  object Label {
+    implicit def encoder: Encoder[Label] =
+      Encoder
+        .forProduct3("add", "remove", "remove_all") { (l: Label) =>
+          (l.add, l.remove, l.removeAll)
+        }
+        .mapJson(l => Json.obj("label" -> l))
+  }
+
+  private[this] object Dummy extends MergifyAction
```
armanbilge, line 63: This trick lets us keep `MergifyAction` `sealed` while being able to compatibly add additional cases to the ADT, since users will never be able to exhaustively pattern match.

armanbilge:
Now, the question is how can I get mergify working on this repo 🤔

.mergify.yml
```diff
@@ -0,0 +1,146 @@
...
+  conditions:
+  - files~=^core
+  actions:
+    label:
+      add:
+      - core
+      remove: []
+- name: Label github PRs
+  conditions:
+  - files~=^github
```
armanbilge, line 53: Hmm, this should end in `/`, or else it will pick up changes in `github-actions` as well.

armanbilge:
This is a whopper, but the generated mergify config is valid (correct, idk 🙃). Once it's out in the wild we'll understand better how it works. I'll keep it stable for 0.4 but if necessary will change some stuff around in 0.5.

nafg:
You could have just used sbt-mergify-github-actions from github.com/nafg/mergify-yaml/

nafg:
It autogenerates the case classes from the Mergify docs so it's complete

armanbilge:
@nafg cool! The hardest part while working on this one was expanding the matrix.
https://github.com/typelevel/sbt-typelevel/blob/9843119f4a2aaaa67e646dd9911db23b9f3a44af/mergify/src/main/scala/org/typelevel/sbt/mergify/MergifyPlugin.scala#L128-L141

Would appreciate any PRs to help with the `case class`es ❤️

nafg:
Not sure I follow what you mean about expanding the matrix.

But you could use mergify-writer as standalone library if you like. Like I said the case classes are autogenerated from Mergify docs.

Or you could submit a PR improving the logic to my SBT plugin ;)

Or if it's explained to me (and it doesn't already do it) then I could add it myself.

joroKr21:
How do I set it up? It looks like just adding the config is not enough. I also tried to install it from here: https://github.com/apps/mergify/installations/new but it's not working: https://github.com/typelevel/cats-tagless/pull/336

PR #192, opened 2022-03-01
Respect project-level `tlVersionIntroduced`

armanbilge:
Fixes a regression introduced in https://github.com/typelevel/sbt-typelevel/pull/176 where `tlVersionIntroduced` was no longer respected on the project-level (it was only using `ThisBuild`-level setting).


armanbilge:
Fixed the issue discovered in https://github.com/typelevel/sbt-typelevel/pull/190.

Issue #202, opened 2022-03-07
A `ci` command alias auto-derived from workflow steps


armanbilge:
As @djspiewak says:
> basically `ci` isn't needed all that often, but when you do need it, you _really_ need it
> at least I've never worked on any long-lived project at any point in my career that didn't, at some point, need it

@rossabaker is this relevant at all e.g. when http4s does a release to patch a CVE? IIRC those are done manually, but is there a reason they can't/aren't done from CI?

Ross also had an idea in https://github.com/typelevel/sbt-typelevel/pull/65#issuecomment-1008422048 to make `ci` a task rather than a command.

I pointed out in https://github.com/typelevel/sbt-typelevel/pull/65#issuecomment-1008426170 that it's difficult/impossible to manipulate an opaque `ci` command and punted the issue. Early milestones of sbt-typelevel worked like this and it was annoying being unable to add a step to `ci` without redefining the entire command.

In any case, since the source-of-truth for CI is the workflow, and the plugin knows the workflow, we should be able to derive the `ci` command automatically. I've definitely thought about this but I'm not sure I have the sbt-fu to pull this off myself 😅

One thing that seems tricky to me is how this would interact with additional CI jobs added to the workflow e.g. as suggested in https://github.com/typelevel/sbt-typelevel/issues/93.

rossabaker:
CVE releases are done manually so a binary is available in Maven Central before the vulnerability is publicly disclosed.

They are a bit dangerous because it's easy to forget a step, particularly MiMa.  I suppose this is why we'd want a `ci` that is reasonably authentic.

armanbilge:
What I don't understand is why you can't just do the release from CI in the private fork that's used to work on the issue?

rossabaker:
CI doesn't run on the private fork.

armanbilge:
Oh, I didn't realize that. Yeesh, so you really really do need a `ci` command because otherwise nobody is checking ...

armanbilge:
TIL about https://github.com/nektos/act for running GHA workflows locally. It has a Nix-thingy therefore Ross will like it 😝

PR #203, opened 2022-03-09
Proposal: Add `mimaReportBinaryIssues` to the `prePR` command

danicheg:
I noticed that `prePR` doesn't contain the `mimaReportBinaryIssues`. Is it intentional? I want to promote using the `prePR` in several projects, so binary compatibility checking is an important step.

### DIFF
--- a/core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
+++ b/core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
@@ -91,6 +91,7 @@ object TypelevelPlugin extends AutoPlugin {
         "scalafmtSbt",
         "set ThisBuild / tlFatalWarnings := tlFatalWarningsInCi.value",
         "Test / compile",
+        "mimaReportBinaryIssues",
         "reload"
       )
     )
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -35,7 +35,7 @@ ThisBuild / tlVersionIntroduced := Map("3" -> "0.4.2")
 
 ## How do I locally prepare my PR for CI?
 
-**sbt-typelevel** comes with a `prePR` command, which updates the GitHub workflow, generates headers, runs `scalafmt`, and clean compiles your code.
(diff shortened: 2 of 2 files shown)


armanbilge:
Thanks for the PR. I agree, I would also like `prePR` to be as helpful as possible.

I think my biggest reservation about `mimaReportBinaryIssues` is that it requires downloading lots and lots of old artifacts. Maybe this is not a big deal, not sure.

The fact that it currently involves a clean recompile is also something I've felt a bit weird about. In fact, I frequently use the `tlPrePrBotHook` command designed for steward since it's much lighter-weight :)

Would be really great to get feedback on this from others!

danicheg:
> I think my biggest reservation about mimaReportBinaryIssues is that it requires downloading lots and lots of old artifacts. Maybe this is not a big deal, not sure.

If `prePR` is supposed to be used actually before submitting the PR, I believe it's worth the hassle. It just saves some time for getting feedback (instead of waiting for the CI and having several iterations).

> The fact that it currently involves a clean recompile is also something I've felt a bit weird about.

I feel that the `prePR` command should contain almost all the steps that the CI job does. In the case of the MiMa tool, it may give some good tips to contributors before submitting the PR. But yeah, it's always a compromise between time and completeness.

danicheg:
Btw one more thing I could advise is to get rid of setting the root project at the initial of the `prePR` command. It would be wasteful for those who made changes in only one submodule.

armanbilge:
> I feel that the `prePR` command should contain almost all the steps that the CI job does.

I think we can instead offer a `ci` command which is a more faithful reproduction, see https://github.com/typelevel/sbt-typelevel/issues/202.

> I believe it's worth the hassle

Idk, I feel really bad about downloading all those artifacts onto contributors locals.

> Btw one more thing I could advise is to get rid of setting the root project at the initial of the prePR command.

Oh yeah, that's probably a good idea.

armanbilge:
Btw, I asked for feedback on Discord #general, in case you want to have a look.

rossabaker:
That sounds fine, but note that some scalafix is "polish" (e.g., organize imports), and some "break something" (e.g., banning Sync stream compiler).

Issue #194, opened 2022-03-02
Document how to setup Sonatype and GPG keys


armanbilge:
This is probably the last important thing missing from our docs. Since we're pushing the passphrase-less keys, we really need our own documentation here.

Solving https://github.com/typelevel/sbt-typelevel/issues/22 would help with that actually. Then we just need to explain the Sonatype part.

Issue #22, opened 2021-12-30
Workflow for setting up signing keys in CI


armanbilge:
This is one from @ChristopherDavenport :)

My idea of how it could work:
> Currently, sbt-gh-actions generates a clean workflow and a CI workflow. It could also generate a third workflow, run by a "dispatch trigger" (i.e., you have to click a button and provide some inputs in web UI) that could:
>
> 1. generate a key
> 2. set it as a repo secret
> 3. send it to a key server
>
> This would make it extremely easy.

https://github.com/typelevel/sbt-typelevel/discussions/11#discussioncomment-1885447 for some discussion

armanbilge:
We now have an org-wide signing key, so this is less of a priority.

PR #206, opened 2022-03-11
Add documentation on how to obtain secrets for publishing

armanbilge:
Closes https://github.com/typelevel/sbt-typelevel/issues/194. And also closes https://github.com/typelevel/sbt-typelevel/issues/22!

Instead of providing a CI workflow for generating keys, we provide a form to do it completely in the browser with OpenPGP.js. I reckon that's just as good.

Screen Shot 2022-03-11 at 14 47 01

### DIFF
--- a/docs/directory.conf
+++ b/docs/directory.conf
@@ -1,6 +1,7 @@
 laika.navigationOrder = [
   index.md
   site.md
+  secrets.md
   faq.md
   customization.md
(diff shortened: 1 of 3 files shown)


armanbilge:
Merging, because docs are good and don't really break stuff when you make a mistake. But always happy for post-facto feedback :)

PR #207, opened 2022-03-14
Port sbt project naming macro for tlCrossRootProject

### LINKED ISSUE #169: Implement `tlCrossRootProject` as macro
In sbt we are used to a project deriving its id from the `val` it is assigned to. Misleadingly, `tlCrossRootProject` currently does not do this. so e.g. if you are to do
```scala
lazy val discipline = tlCrossRootProject.aggregate(core)
```
your root project is still called `root`.

An example of how to implement this with a macro is in sbt-crossproject, should not be too hard to adapt.
https://github.com/portable-scala/sbt-crossproject/blob/829fd118205e20973f8751540836f3b902e86623/sbt-crossproject/src/main/scala/sbtcrossproject/CrossProjectMacros.scala

cc @vasilmkd

DavidGregory084:
This PR ports the logic of the sbt `project` macro to ensure that projects are named after their enclosing `val`. Resolves #169.

From what I can tell this seems to work on sbt-typelevel itself, e.g. if I rename `lazy val root`:
image

or fail to assign it:

image

Contributed on behalf of the @opencastsoftware open source team. 👋

Is this the correct base branch @armanbilge?


.editorconfig
```diff
@@ -0,0 +1,11 @@
...
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.{scala,sbt}]
+indent_style = space
+indent_size = 2
```
DavidGregory084, line 11: I generally add this to most projects as it makes editors that understand it behave themselves w.r.t. indentation - hope you don't mind

ci/src/main/scala/org/typelevel/sbt/CrossRootProject.scala
```diff
@@ -99,11 +99,13 @@ object CrossRootProject {
...
-    Project("rootNative", file(".native"))
+  def apply(): CrossRootProject = CrossRootProject("root")
+
+  def apply(name: String): CrossRootProject = new CrossRootProject(
+    Project(name, file(".")),
+    Project(s"${name}JVM", file(".jvm")),
+    Project(s"${name}JS", file(".js")),
+    Project(s"${name}Native", file(".native"))
   ).enablePlugins(NoPublishPlugin, TypelevelCiCrossPlugin)
 }
```
DavidGregory084, line 110: These must become public as the call will be inlined by the macro

ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
```diff
@@ -21,14 +21,15 @@ import org.typelevel.sbt.gha.GenerativePlugin
...
+import scala.language.experimental.macros
 
 object TypelevelCiPlugin extends AutoPlugin {
 
   override def requires = GitHubActionsPlugin && GenerativePlugin && MimaPlugin
   override def trigger = allRequirements
 
   object autoImport {
-    def tlCrossRootProject: CrossRootProject = CrossRootProject()
+    def tlCrossRootProject: CrossRootProject = macro CrossRootProjectMacros.crossRootProjectImpl
```
DavidGregory084, line 32: Note that this causes a MiMa issue as this `def` technically no longer exists

DavidGregory084: Should I add an exclusion or do you know of a good workaround for problems like this?

armanbilge:
Wow, this is fantastic, thank you so much! 😁 

> Is this the correct base branch @armanbilge?

Good question! I was wondering that myself. I'd love to get this into `series/0.4` but this change is probably just a little too breaking, since for some users the id of their root project will suddenly change. And also the MiMa issue, which I'm not sure if we can workaround.

Let me get `main` up-to-date with `series/0.4` and then we can retarget your PR.

.editorconfig
armanbilge: Whatever it takes to keep the OSS contributions coming 😝

armanbilge:
Re-triggering CI.

ci/src/main/scala/org/typelevel/sbt/CrossRootProject.scala
```diff
@@ -99,11 +99,13 @@ object CrossRootProject {
...
     Some((root.all, root.jvm, root.js, root.native))
 
-  private[sbt] def apply(): CrossRootProject = new CrossRootProject(
-    Project("root", file(".")),
-    Project("rootJVM", file(".jvm")),
-    Project("rootJS", file(".js")),
-    Project("rootNative", file(".native"))
+  def apply(): CrossRootProject = CrossRootProject("root")
+
+  def apply(name: String): CrossRootProject = new CrossRootProject(
```
armanbilge, line 104: Nit, it's technically the `id` :)

```suggestion
  def apply(id: String): CrossRootProject = new CrossRootProject(
```

armanbilge (commented):
These changes are looking good! 😁 

Now, the tricky part is that we need to make sure to fix all the places where the "root" assumption is baked in. For example:
https://github.com/typelevel/sbt-typelevel/blob/b1bbae5bfb99bf0b0e04980b904627ca05e19cd0/ci/src/main/scala/org/typelevel/sbt/CrossRootProject.scala#L132

There's definitely a few strategies we can take. Off the top of my head:
1. We can probably access the `id` setting of the `LocalRootProject`.
2. We can make those `AutoPlugin`s into `class`es which take an `id` as an argument.

WDYT?

DavidGregory084:
> These changes are looking good! 😁
> 
> Now, the tricky part is that we need to make sure to fix all the places where the "root" assumption is baked in. For example:
> 
> https://github.com/typelevel/sbt-typelevel/blob/b1bbae5bfb99bf0b0e04980b904627ca05e19cd0/ci/src/main/scala/org/typelevel/sbt/CrossRootProject.scala#L132
> 
> There's definitely a few strategies we can take. Off the top of my head:
> 
>     1. We can probably access the `id` setting of the `LocalRootProject`.
> 
>     2. We can make those `AutoPlugin`s into `class`es which take an `id` as an argument.
> 
> 
> WDYT?

Hmm that's a tricky one - I have not tried (or seen anyone try) making an AutoPlugin that is not an `object` before. Would that cause problems with automatic inclusion like `requires`?

I think that 1. is probably the best bet - we might be able to get the `state.value` and then get `Project` details using `Project.extract(state)`?

armanbilge:
> Hmm that's a tricky one - I have not tried (or seen anyone try) making an AutoPlugin that is not an `object` before. Would that cause problems with automatic inclusion like `requires`?

Yes, I've never seen this done before either. These plugins are only meant to be used internally to support the `CrossRootProject` functionality so `requires` is not a concern, but other weird stuff could happen, for sure :)
https://github.com/typelevel/sbt-typelevel/blob/a4282aed116872daa6b9cd77c40804ceacc73f6f/ci/src/main/scala/org/typelevel/sbt/CrossRootProject.scala#L85-L86

> I think that 1. is probably the best bet - we might be able to get the `state.value` and then get `Project` details using `Project.extract(state)`?

Ok, 👍 let's give that a try!

DavidGregory084:
OK, I think I have removed any hardcoded project IDs now @armanbilge. You were of course right about `LocalRootProject` - never seen that one before! I had forgotten that `State` can't be used in `Setting`s.

ci/src/main/scala/org/typelevel/sbt/CrossRootProject.scala
```diff
@@ -20,8 +20,8 @@ import sbt._
 import org.typelevel.sbt.gha.GenerativePlugin.autoImport._
 
 /**
- * Simultaneously creates a `root`, `rootJVM`, `rootJS`, and `rootNative` project, and
- * automatically enables the `NoPublishPlugin`.
+ * Simultaneously creates a root project, a Scala JVM subproject, a Scala.js subproject, a Scala
+ * Native subproject and automatically enables the `NoPublishPlugin`.
```
armanbilge, line 24: I actually feel like this is less clear about what it does :)

Maybe "aggregate project" would be clearer than "sub project"?

ci/src/main/scala/org/typelevel/sbt/CrossRootProject.scala
```diff
@@ -96,14 +96,16 @@ final class CrossRootProject private (
...
-
-  private[sbt] def apply(): CrossRootProject = new CrossRootProject(
-    Project("root", file(".")),
-    Project("rootJVM", file(".jvm")),
-    Project("rootJS", file(".js")),
-    Project("rootNative", file(".native"))
+  def unapply(rootProject: CrossRootProject): Some[(Project, Project, Project, Project)] =
+    Some((rootProject.all, rootProject.jvm, rootProject.js, rootProject.native))
+
+  def apply(): CrossRootProject = CrossRootProject("root")
```
armanbilge, line 102: Eh, I'm not so crazy about a default constructor. Since it was never public we can just remove it.

ci/src/main/scala/org/typelevel/sbt/CrossRootProject.scala
```diff
@@ -126,8 +128,10 @@ object TypelevelCiJVMPlugin extends AutoPlugin {
   override def requires = TypelevelCiCrossPlugin
 
   override def buildSettings: Seq[Setting[_]] = Seq(
-    githubWorkflowBuildMatrixAdditions ~= { matrix =>
-      matrix.updated("project", matrix("project") ::: "rootJVM" :: Nil)
+    githubWorkflowBuildMatrixAdditions := {
+      val matrix = githubWorkflowBuildMatrixAdditions.value
+      val rootProject = (LocalRootProject / Keys.thisProject).value
```
armanbilge, line 133: To save some repetition, we could extract `rootProjectId` into a `Def.setting` maybe.

armanbilge (commented):
Thanks! Just a few nits.

Since I'm terribly lazy with tests I try to dog-food as much as possible. In fact, that's why the root is a `tlCrossRootProject` even though this plugin is obviously JVM-only.

https://github.com/typelevel/sbt-typelevel/blob/ae333f666f9964718095bc2d27184b01a8e77465/build.sbt#L18

Maybe let's rename our root like this and see if anything breaks 😆 
```scala
lazy val `sbt-typelevel` = tlCrossRootProject.aggregate(
```

ci/src/main/scala/org/typelevel/sbt/CrossRootProject.scala
DavidGregory084: Hmm I thought about doing so, but once you have declared a SettingKey in your plugin it kind of becomes part of the public API of that plugin across every scope axis - do you think it's worth it? Totally happy to do so if you prefer that!

DavidGregory084:
Sorry wasn't ignoring you @armanbilge - was just onboarding some new folks end of last week 😛 

I think I have addressed the comments here now except for one question

ci/src/main/scala/org/typelevel/sbt/CrossRootProject.scala
armanbilge: Hmm, I might be learning something here, but isn't it possible to have anonymous settings? I hope so because I used them everywhere 😆 For example like this one, what's the setting key for it?
https://github.com/typelevel/sbt-typelevel/blob/50e6ddfb93ff143c293087d5b8159738a0f3363f/github/src/main/scala/org/typelevel/sbt/TypelevelGitHubPlugin.scala#L58

armanbilge:
No worries, you've been doing really fantastic work!! 😁 And ha ha, I am super-duper pleased to see that it's happily dog-fooding itself 🚀 very very nice.

ci/src/main/scala/org/typelevel/sbt/CrossRootProject.scala
DavidGregory084: Oh no it's definitely me that's learning - that's a nice pattern that I hadn't seen before

DavidGregory084: OK applied that one as well now 🚀

armanbilge (approved):
LGTM! Thank you very much for all of your work on this :)

PR #225, opened 2022-03-24
Add the API base URI to the Laika config

DavidGregory084:
This enables the [API Documentation Linking](https://planet42.github.io/Laika/0.18/03-preparing-content/02-navigation.html#linking-to-api-documentation) feature of Laika to work if API doc base URL is declared in the build config.

### DIFF
--- a/site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
+++ b/site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
@@ -23,6 +23,8 @@ import laika.helium.config.Favicon
 import laika.helium.config.HeliumIcon
 import laika.helium.config.IconLink
 import laika.helium.config.ImageLink
+import laika.rewrite.link.ApiLinks
+import laika.rewrite.link.LinkConfig
 import laika.sbt.LaikaPlugin
 import laika.theme.ThemeProvider
 import mdoc.MdocPlugin
@@ -105,6 +107,12 @@ object TypelevelSitePlugin extends AutoPlugin {
       .value: @nowarn("cat=other-pure-statement"),
     tlSitePreview := previewTask.value,
     Laika / sourceDirectories := Seq(mdocOut.value),
+    laikaConfig := {
+      val currentConfig = laikaConfig.value
+      tlSiteApiUrl.value.fold(currentConfig) { apiUrl =>
+        currentConfig.withConfigValue(LinkConfig(apiLinks = Seq(ApiLinks(apiUrl.toString))))
+      }
+    },
     laikaTheme := tlSiteHeliumConfig.value.build.extend(tlSiteHeliumExtensions.value),
     mdocVariables := {
       mdocVariables.value ++


armanbilge:
Aha, very nice! I was recently experimenting with this feature in [http4s-dom](https://github.com/http4s/http4s-dom/blob/af4d02aa4d93ff692504132407ccdbf2eee00889/build.sbt#L179).

I really want to do something like this to. There's a couple bits of trickyness:
1. The `tlSiteApiUrl` is not necessarily the form appropriate base URL for the Laika's API linking feature ... although this may have been a problem of my own making 😅 see:
https://github.com/typelevel/sbt-typelevel/blob/f2db840015c44621f15607eb40b0cd9abbd9e829/site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala#L137-L138
Practically, it's the difference between:
    - https://www.javadoc.io/doc/org.typelevel/sbt-typelevel-docs_2.12/latest/index.html
    - https://www.javadoc.io/doc/org.typelevel/sbt-typelevel-docs_2.12/latest/org/typelevel/sbt/index.html
 
    In fact, I reckon if you try your change with the sbt-typelevel docs it'll have this problem. (This is the basic principle of sbt-typelevel: dog-food everything!)

2. In user-land, if I want to add more packages to this `apiLinks` config, AFAICT it doesn't really compose well and I'd have to start from scratch. If we exposed this specific config as a sbt setting, this would make it easier.

    I'm not sure if this is a common enough problem to justify building the machinery for it, but it is helpful for http4s where the satellite repos tend to be implementations of APIs living in the core repo. Also it would be really cool to have this work out of the box for all cats(-effect) APIs etc.

WDYT?

DavidGregory084:
>     1. The `tlSiteApiUrl` is not necessarily the appropriate form of base URL for the Laika's API linking feature ... although this may have been a problem of my own making 😅 see:
>        https://github.com/typelevel/sbt-typelevel/blob/f2db840015c44621f15607eb40b0cd9abbd9e829/site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala#L137-L138
>        
>        Practically, it's the difference between:
>        
>        * https://www.javadoc.io/doc/org.typelevel/sbt-typelevel-docs_2.12/latest/index.html
>        * https://www.javadoc.io/doc/org.typelevel/sbt-typelevel-docs_2.12/latest/org/typelevel/sbt/index.html
>        
>        In fact, I reckon if you try your change with the sbt-typelevel docs it'll have this problem. (This is the basic principle of sbt-typelevel: dog-food everything!)
Hmm yes I see what you mean; I was testing this with scalacache but I neglected to set the `tlSiteApiPackage` so everything looked correct. I suppose we could move the usage of `tlSiteApiPackage` out of the `tlSiteApiUrl` and into the code which creates the API doc links? It's still a nice usability feature.

>     2. In user-land, if I want to add more packages to this `apiLinks` config, AFAICT it doesn't really compose well and I'd have to start from scratch. If we exposed this specific config as a sbt setting, this would make it easier.
>        I'm not sure if this is a common enough problem to justify building the machinery for it, but it is helpful for http4s where the satellite repos tend to be implementations of APIs living in the core repo. Also it would be really cool to have this work out of the box for all cats(-effect) APIs etc.
Yeah it seems the Laika config is a little tricky to work with in general, although I think that `withConfigValue` might be additive as it seems updates an underlying builder object? It's a shame that `=~` doesn't work well with `.value` anymore as it would make this kind of thing easier to use.

site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
```diff
@@ -105,6 +107,12 @@ object TypelevelSitePlugin extends AutoPlugin {
       .value: @nowarn("cat=other-pure-statement"),
     tlSitePreview := previewTask.value,
     Laika / sourceDirectories := Seq(mdocOut.value),
+    laikaConfig := {
+      val currentConfig = laikaConfig.value
+      tlSiteApiUrl.value.fold(currentConfig) { apiUrl =>
+        currentConfig.withConfigValue(LinkConfig(apiLinks = Seq(ApiLinks(apiUrl.toString))))
```
armanbilge, line 113: Following on the additivity issue, `.withConfigValue` is definitely additive. But I assume that this will override any previously set `LinkConfig` and esp. all `apiLinks` previously set?

DavidGregory084: I guess only way is to test it!
```scala
scala> import laika.sbt._
import laika.sbt._

scala> import laika.rewrite.link._
import laika.rewrite.link._

scala> val conf = LaikaConfig.defaults
conf: laika.sbt.LaikaConfig = LaikaConfig(UTF-8,<function1>,laika.config.ConfigBuilder@302e25d3,laika.ast.MessageFilter$$anon$1@78ea076,None,laika.ast.MessageFilter$$anon$1@56276b5d)

scala> val confWithLinks = conf.withConfigValue(LinkConfig(apiLinks = List(ApiLinks(baseUri = "https://www.javadoc.io/doc/com.github.cb372/scalacache-unidocs_2.13/0.28.0/")))
     | )
confWithLinks: laika.sbt.LaikaConfig = LaikaConfig(UTF-8,<function1>,laika.config.ConfigBuilder@623f79d4,laika.ast.MessageFilter$$anon$1@78ea076,None,laika.ast.MessageFilter$$anon$1@56276b5d)
                                                                                                                                                                                                           
scala> val confWithMoreLinks = confWithLinks.withConfigValue(LinkConfig(apiLinks = List(ApiLinks(baseUri = "https://www.javadoc.io/doc/org.typelevel/sbt-typelevel-docs_2.12/latest/", "org.typelevel.sbt"))))
confWithMoreLinks: laika.sbt.LaikaConfig = LaikaConfig(UTF-8,<function1>,laika.config.ConfigBuilder@1698fd02,laika.ast.MessageFilter$$anon$1@78ea076,None,laika.ast.MessageFilter$$anon$1@56276b5d)

scala> confWithLinks.configBuilder.build.get[LinkConfig]("laika.links")
res0: laika.config.Config.ConfigResult[laika.rewrite.link.LinkConfig] = Right(LinkConfig(List(),List(),List(ApiLinks(https://www.javadoc.io/doc/com.github.cb372/scalacache-unidocs_2.13/0.28.0/,*,index.html)),List()))

scala> confWithMoreLinks.configBuilder.build.get[LinkConfig]("laika.links")
res1: laika.config.Config.ConfigResult[laika.rewrite.link.LinkConfig] = Right(LinkConfig(List(),List(),List(ApiLinks(https://www.javadoc.io/doc/org.typelevel/sbt-typelevel-docs_2.12/latest/,org.typelevel.sbt,index.html)),List()))

scala>
```
It looks like yes, latest `ApiLinks` wins 😬

armanbilge:
> I suppose we could move the usage of `tlSiteApiPackage` out of the `tlSiteApiUrl` and into the code which creates the API doc links? It's still a nice usability feature.

Yeah, although it's less about the usage of `tlSiteApiPackage` and more that `tlSiteApiUrl` currently means "what the API docs button in the website links to" and not "a machine-readable API url". Presumably anyone can set `tlSiteApiUrl` to anything :)

> Yeah it seems the Laika config is a little tricky to work with in general

Yeppp. For example, right now I don't think it's possible to modify the config to replace/remove the Typelevel logo and social links at the top without losing the API and github links as well. This is unfortunate for the http4s plugin which has to replicate this functionality and for anyone outside Typelevel that wants to enjoy the goodies in this plugin.

https://github.com/http4s/sbt-http4s-org/blob/e2c730ccedc3fd13495fe2ea0c38eaad2e4bbdce/core/src/main/scala/org/http4s/sbt/Http4sOrgSitePlugin.scala#L93

So, our recourse is to expose all these things in sbt settings, as I'm tempted to here. Then we can auto-populate it with useful stuff (like the projects own API docs) while letting downstreams add whatever else they might need.

armanbilge:
See also:
- https://github.com/planet42/Laika/issues/285

PR #240, opened 2022-04-03
Demonstrate how to add unidoc project filter in docs

armanbilge:
And make the example project a bit more fun 🍯

### DIFF
--- a/docs/site.md
+++ b/docs/site.md
@@ -39,7 +39,8 @@ lazy val unidocs = project
   .in(file("unidocs"))
   .enablePlugins(TypelevelUnidocPlugin) // also enables the ScalaUnidocPlugin
   .settings(
-    name := "woozle-docs"
+    name := "woozle-docs",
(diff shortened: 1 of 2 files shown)


docs/site.md
```diff
@@ -39,7 +39,8 @@ lazy val unidocs = project
   .in(file("unidocs"))
   .enablePlugins(TypelevelUnidocPlugin) // also enables the ScalaUnidocPlugin
   .settings(
-    name := "woozle-docs"
+    name := "woozle-docs",
+    ScalaUnidoc / unidoc / unidocProjectFilter := inProjects(core.jvm, heffalump)
```
armanbilge, line 43: I'd still like to think about ways to automate this setting. It's tricky to do in general—you can't just ignore JS/native projects, because some modules may only be published for those.

armanbilge: Btw @timwspence I think adding a setting like this should fix the cats-stm build.

Issue #138, opened 2022-02-03
Support for `-Yscala-release` flag


armanbilge:
This is the new flag in Scala 3.1.2-RC1 that enables targeting Scala 3.0 from the Scala 3.1 compiler. One of the exciting things about sbt-typelevel is that we can remotely roll out this feature to several projects :)

I've been thinking about this, and I propose we add a `tlScala3Release` setting to the `typelevel-settings` plugin that defaults to 3.0.

@smarter some projects such as fs2 are already being published against Scala 3.1. Is it okay to target them to Scala 3.0 going forward? Otherwise those projects will have to overwrite this setting.

It also seems there's some fiddling required with POM files. @smarter could you clarify what exactly this is? Thanks!

Seems related to:
```scala
dependencyOverrides ++= Seq(
  scalaOrganization.value %% "scala3-library" % scalaVersion.value,
  scalaOrganization.value %% "scala3-library_sjs1" % scalaVersion.value // for Scala.js projects
)
```
https://docs.scala-lang.org/scala3/reference/language-versions/binary-compatibility.html

smarter:
First I'd like to insist on the fact that -Yscala-release is an **experimental** feature (hence the use of -Y, the flag will be renamed to -scala-release when it's deemed stable), it shouldn't be used to publish non-experimental artifacts and it certainly shouldn't be used by default in an sbt plugin.

> Is it okay to target them to Scala 3.0 going forward?

Not in general as mentioned in https://github.com/typelevel/cats/pull/4016#issuecomment-947750254
Also, I think it's up to each project to decide what minimal version of Scala they want to support depending on what features/bugfixes they rely on. If you use -Yscala-release 3.0 you should consider actually testing that your APIs are properly usable from 3.0 (e.g., to make sure there's no inference issues).

> It also seems there's some fiddling required with POM files. @smarter could you clarify what exactly this is?

I don't have more details than what's in https://docs.scala-lang.org/scala3/reference/language-versions/binary-compatibility.html, right now sbt doesn't know anything about the flag, so when you're using `scalaVersion := 3.1.2-RC1, scalacOptions ++= Seq("-Yscala-release", "3.0")`, sbt will declare the 3.1.2-RC1 version of the scala-library as a pom dependency even though in practice 3.0 would be a valid dependency. The plan is to fix this in sbt somehow before -scala-release becomes non-experimental.

armanbilge:
[...]

PR #241, opened 2022-04-04
add scala 3 release version flag

i10416:


i10416:
rel https://github.com/typelevel/sbt-typelevel/issues/138

armanbilge:
Thanks for the PR! :) A few quick notes:

- Actually the compiler flag was renamed to `-scala-output-version` :)
- Unfortunately, adding the compiler flag is not sufficient; we also need to modify the POM to use the old Scala standard library instead of thew one. Fortunately, sbt 1.7.0 will implement the extra functionality, so we can directly integrate with their setting.
    https://github.com/sbt/sbt/releases/tag/v1.7.0-M1
- I think we should make the default Scala 3.1.1

Also, would you mind targeting your PR at `main` branch instead of `series/0.4`? We will release this feature in sbt-typelevel v0.5.0 :)

i10416:
Hmm, should we wrap `scalaOutputVersion := "3.0.2"` in `tlScalaRelease` rather than exposing it to users?

armanbilge:
That's a good question. I don't think we need our own `tlScalaRelease` setting. I'm not sure what we would use it for.

i10416:
Sure.

In addition, are we going to use `scalaOutputVersion` here, which is still pre-release , or wait for next sbt release?. What do you think?

armanbilge:
I think it's too hard to support this Scala 3 feature without relying on the sbt integration. So, we definitely need to rely on that feature.

I can't promise I'll merge this PR right away. I have to double-check when the planned release date for sbt v1.7.0 is and and decide whether sbt-typelevel v0.5.0 should wait for it or not. I don't currently feel a big rush to release sbt-typelevel v0.5.0 so that strategy may make sense.

Meanwhile I am publishing some milestone prereleases of sbt-typelevel v0.5.0 so some people can use that with the sbt milestone prerelease, if they want.

i10416:
Okay. I'm not in a hurry too. 
We'd better waiting for sbt 1.7.0 release for v0.5.0. I close this PR for now so that someone desperately needs that feature can feel free to open PR.

i10416:
@armanbilge Thanks for reviewing :)

armanbilge:
Thanks for your contribution! :)

armanbilge:
@i10416 if you're still interested in working on this one, I had an idea for how we can do this so it will be ready before sbt 1.7.0 is released, and also so we don't have to bump to sbt-typelevel 0.5.0.

Basically, the idea is to use sbt "reflection", when you refer to a setting using a `String`. Here's an example:
https://github.com/typelevel/sbt-typelevel/blob/66b313c7bdd7c6a8f35b5d6c41c9c4f9a19a773b/sonatype/src/main/scala/org/typelevel/sbt/TypelevelUnidocPlugin.scala#L38
This technique will work even if somebody uses sbt-typelevel with sbt 1.6.x. It won't _do_ anything, but at least it won't crash the build :)

i10416:
@armanbilge  Interesting! I didn't know that technique.

Issue #215, opened 2022-03-18
Java 8 in CI: the end is nigh


armanbilge:
Maybe not tomorrow or even this year, but this stuckfast position will gradually grow more and more uncomfortable. E.g. our sbt-plugins are itching to leave us behind:
- https://github.com/sbt/sbt-git/issues/213

FWIW we can probably get along just fine without JDK 8 in CI. But perhaps I haven't heard the horror stories 😆 
1. The `tlJdkRelease := Some(8)` setting enables actives all the scalac and javac flags to safely build/release from a newer JDK.
2. fs2 only runs CI on JDK 17 and besides the occasional bump is really a success story. Note that the `tlJdkRelease` setting would actually prevent those `ByteBuffer`-oopsies. In fact fs2 _can't_ use the `tlJdkRelease` setting because it blocks access to JDK 9+ APIs!

rossabaker:
Java 8 Active Support ended on March 31, which is historically the date I become less of an curmudgeon about maintaining old things.

armanbilge:
Politely pinging @vasilmkd, do you have any opinions on this? Things are still a bit new with the `tlJdkRelease` flag on Scala 3 but actually so far it seems to be doing _better_ than vanilla JDK 8 🤣

vasilmkd:
I wouldn't completely drop it in Cats Effect (and we would set that as a special case), but as a default, I think we should be nudging users towards the future and make JDK 17 the default.

armanbilge:
Yup, this is definitely where things are headed. /cc @valencik
- https://github.com/scalameta/metals/pull/3788

In v0.5.0 we'll set `tlJdkRelease := 8`, switch the default CI JDK to 17, and upgrade the nix flake in the g8 template as well.

armanbilge:
@bpholt thanks, that's helpful input. Projects such as Cats, Cats Effect, and Http4s will almost certainly want to keep the Java 8 builds as well.

So I guess we have to be careful to make sure sbt-typelevel still works on JDK 8, even if we switch the default CI to JDK 17 and set the release flag to Java 8 as proposed here.

We can keep thinking about this as we approach sbt-typelevel v0.5.x.

armanbilge:
Another interesting idea:

So sbt has a setting for running scalac and forked tests on a different JDK than what sbt itself is running on (h/t @keynmol).
[...]

PR #243, opened 2022-04-05
Use JDK 17 for CI in sbt-typelevel plugin

armanbilge:
Closes https://github.com/typelevel/sbt-typelevel/issues/215.

We can only safely make this change in the core, opinionated plugin, because that plugin also sets the `tlJdkRelease` setting appropriately.


armanbilge:
I don't think this is happening this time around 😕

PR #244, opened 2022-04-06
Use SJS linker batch mode in CI

armanbilge:
According to scaladoc:

> In batch mode, the linker can throw away intermediate state that is otherwise maintained for incremental runs.

No idea if it will make a meaningful difference, but we definitely don't need incremental state in CI.


github/src/main/scala/org/typelevel/sbt/TypelevelScalaJSGitHubPlugin.scala
```diff
@@ -43,6 +44,9 @@ object TypelevelScalaJSGitHubPlugin extends AutoPlugin {
           s"$flag$l->$g"
         }
       }
+    },
+    scalaJSLinkerConfig := {
+      scalaJSLinkerConfig.value.withBatchMode(sys.env.get("GITHUB_ACTIONS").contains("true"))
```
armanbilge, line 49: This plugin annoyingly doesn't know about the GH actions plugin. So it has to check for itself whether we are in CI.

Issue #253, opened 2022-04-11
Flag `-Werror` set repeatedly on `0.4.8` version


danicheg:
`0.4.8` contains some regression on compiling Scala 3 sources. 
Related failed CI builds: https://github.com/typelevel/log4cats/runs/5969372003?check_suite_focus=true#step:10:30, https://github.com/typelevel/keypool/runs/5969330317?check_suite_focus=true#step:9:28

armanbilge:
Thanks for the report. I guess this broke everyone's build ...

PR #255, opened 2022-04-11
Revert #250 which moved scalac/javac options to scope-specific settings

armanbilge:
Fixes https://github.com/typelevel/sbt-typelevel/issues/253 by effectively reverting https://github.com/typelevel/sbt-typelevel/pull/250.

I though I was very clever in https://github.com/typelevel/sbt-typelevel/pull/250 but I neglected to realize that the `Test` config inherits from the `Compile` config. So, if you setting scalac and javac options additively, then the settings in the `Test` scope are not independent of those in `Compile`.

sbt-tpolecat recently navigated this in https://github.com/DavidGregory084/sbt-tpolecat/pull/58, and AFAICT was able to do so because its sets scalac options all at once e.g. `:=` instead of `++=`. But that's too big a change for now.

### DIFF
--- a/settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
+++ b/settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
@@ -195,10 +195,9 @@ object TypelevelSettingsPlugin extends AutoPlugin {
       "-encoding",
       "utf8",
       "-Xlint:all"
-    )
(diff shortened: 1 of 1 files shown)


armanbilge:
Hm, interesting note: 2 days after my PR here sbt-tpolecat changed from using `:=` to `++=` in https://github.com/DavidGregory084/sbt-tpolecat/pull/61.

PR #264, opened 2022-04-21
Don't use tags for versioning if git tree is dirty

armanbilge:
Philosophically, if there are uncommitted changes then arguably any tags on the HEAD commit are no longer relevant and should not be used for versioning. (If you made a new commit with the current changes, then they definitely wouldn't be.)

Practically, if your HEAD commit is tagged 0.3.0 and you bump your base version to 0.4 _without_ committing, then attempting to load sbt-typelevel will error with:
```
[error] Your current version 0.3.0 cannot be less than tlBaseVersion 0.4
```
This error is a fail-safe to prevent publishing of an 0.3.0 when the base version is 0.4, since that would be bad. But if your git tree is dirty, then your version will not be v0.3.0 anyway, so this error is pointlessly confusing and annoying.

So, if the git tree is dirty, ignore the tags for the purposes of versioning.


tpolecat (approved):
:+1: Thanks!

Issue #247, opened 2022-04-07
Proposal: move `sbt-scalafmt` to the `TypelevelKernelPlugin`


danicheg:
I think that `sbt-scalafmt` is the base plugin for every Typelevel project. So I opine it's natural to move it to the most upstream `Typevel*` plugin (I guess it's the `TypelevelKernelPlugin`). WDYT?

armanbilge:
Well, not _every_ project :) e.g. https://github.com/typelevel/shapeless-3 and maybe some others I forget.

Also, there are projects outside Typelevel that use the plugin but don't like using scalafmt, such as @davenverse and tpolecat-verse.

Is your proposal just that we add the dependency, or that we also add the CI integration? The `TypelevelKernelPlugin` also doesn't know about GH actions, that's also in a separate plugin 😅

I think the long-term goal should be that all Typelevel projects use `sbt-typelevel` instead of `sbt-typelevel-ci-release`. Then it doesn't matter where the formatting plugin is.

danicheg:
Analogically that client is getting `sbt-mima-plugin` transitively from `sbt-typelevel-ci-release`, it'd be cool to get something the same with `sbt-scalafmt`. Just because the formatting looks like something base.

danicheg:
That was my point - to not use the most powered plugin because I need a little thing (formatting) from it. And it seems possible to move down the formatting plugin within the Typelevel plugins hierarchy.

armanbilge:
Right, b/c there are some things in the powered plugin that you do not like? :) The problem is, there are people who also don't like formatting, and I think it's important for sbt-typelevel to be useful to them as well. Nobody agrees on anything! :)

What I think could make this easier is to extract an `sbt-typelevel-ci-scalafmt` plugin. So then anyone can add it to their build easily. I originally tried this in https://github.com/http4s/sbt-http4s-org/pull/120#issuecomment-1008454875 but I ran into some problems because of sbt non-determinism.

Btw, FTR I would _really_ like to solve this. I just don't know how.

armanbilge:
Ok, I have an idea how to fix this :)

We can use sbt "reflection" in the sbt-typelevel-ci plugin to check if the scalafmt, header, and mima plugins are enabled in the build. If they are, we can add the steps to CI automatically.

Note this means we won't actually add these dependencies; they will still have to be added somewhere else. However, if they are present, then CI will configure itself automatically.

PR #274, opened 2022-05-17
Add Scalafix integration

DavidGregory084:
Currently includes only [liancheng/scalafix-organize-imports](https://github.com/liancheng/scalafix-organize-imports), could be used later to roll out [typelevel-scalafix](https://github.com/typelevel/typelevel-scalafix).

Unfortunately it's not totally straightforward to roll this out as a CI step - e.g. if we consider cats, there are currently some bincompat tricks in place that conflict with the simulacrum-scalafix generated code, so it's currently (deliberately) not in sync with its Scalafix migrations.
Unless we implemented some kind of "redirect" or "migrate" annotation to handle those cases in the simulacrum code by dispatching a deprecated ops method to its replacement, I don't think Github Actions part of this could be enabled in cats by default as it would always produce a diff.

Another concern is that enabling Scalafix in check mode could lead to build failures whenever a Scalafix rule produces code that is not correctly formatted according to `scalafmtCheckAll`. I have asked on the Scalameta Discord about whether it's possible to run Scalafmt on Scalafix output before running checks.

### DIFF
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -1,0 +1,7 @@
+rules = [
+    OrganizeImports
+]
+
+OrganizeImports {
+  preset = INTELLIJ_2020_3
+}
--- a/core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
+++ b/core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
@@ -16,12 +16,13 @@
 
 package org.typelevel.sbt
 
-import sbt._, Keys._
-import org.typelevel.sbt.gha.GenerativePlugin
-import org.typelevel.sbt.gha.GitHubActionsPlugin
 import de.heikoseeberger.sbtheader.HeaderPlugin
+import org.typelevel.sbt.gha.{GenerativePlugin, GitHubActionsPlugin}
+import sbt._
 
 import scala.collection.immutable
+
+import Keys._
 
 object TypelevelPlugin extends AutoPlugin {
 
@@ -44,7 +45,6 @@ object TypelevelPlugin extends AutoPlugin {
   import TypelevelKernelPlugin.mkCommand
   import TypelevelSettingsPlugin.autoImport._
   import TypelevelSonatypeCiReleasePlugin.autoImport._
-  import TypelevelScalafixPlugin.autoImport._
   import GenerativePlugin.autoImport._
   import GitHubActionsPlugin.autoImport._
 
--- a/core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
+++ b/core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
@@ -29,6 +29,7 @@ object TypelevelPlugin extends AutoPlugin {
     TypelevelKernelPlugin &&
       TypelevelSettingsPlugin &&
       TypelevelCiReleasePlugin &&
+      TypelevelScalafixPlugin &&
       GitHubActionsPlugin &&
       HeaderPlugin
 
@@ -43,6 +44,7 @@ object TypelevelPlugin extends AutoPlugin {
   import TypelevelKernelPlugin.mkCommand
   import TypelevelSettingsPlugin.autoImport._
   import TypelevelSonatypeCiReleasePlugin.autoImport._
+  import TypelevelScalafixPlugin.autoImport._
   import GenerativePlugin.autoImport._
   import GitHubActionsPlugin.autoImport._
 
@@ -73,8 +75,8 @@ object TypelevelPlugin extends AutoPlugin {
     },
     githubWorkflowBuild := {
       WorkflowStep.Sbt(
-        List("headerCheckAll", "scalafmtCheckAll", "project /", "scalafmtSbtCheck"),
-        name = Some("Check headers and formatting"),
+        List("headerCheckAll", "scalafmtCheckAll", "scalafixAll --check", "project /", "scalafmtSbtCheck"),
+        name = Some("Check headers, migrations and formatting"),
         cond = Some(primaryJavaCond.value)
       ) +: githubWorkflowBuild.value
     }
--- a/scalafix/src/main/scala/org/typelevel/sbt/TypelevelScalafixPlugin.scala
+++ b/scalafix/src/main/scala/org/typelevel/sbt/TypelevelScalafixPlugin.scala
@@ -1,7 +1,9 @@
 package org.typelevel.sbt
 
-import sbt._, Keys._
+import sbt._
 import scalafix.sbt.ScalafixPlugin
+
+import Keys._
 import ScalafixPlugin.autoImport._
 
 object TypelevelScalafixPlugin extends AutoPlugin {
--- a/scalafix/src/main/scala/org/typelevel/sbt/TypelevelScalafixPlugin.scala
+++ b/scalafix/src/main/scala/org/typelevel/sbt/TypelevelScalafixPlugin.scala
@@ -1,0 +1,28 @@
+package org.typelevel.sbt
+
+import sbt._, Keys._
+import scalafix.sbt.ScalafixPlugin
+import ScalafixPlugin.autoImport._
+
+object TypelevelScalafixPlugin extends AutoPlugin {
+
+  override def requires = ScalafixPlugin
+
+  override def trigger = allRequirements
+
+  object autoImport {
+    val tlScalafixDependencies = settingKey[Seq[ModuleID]]("The scalafix rule dependencies to enable in the build.")
+  }
+
+  import autoImport._
+
+  override def buildSettings = Seq[Setting[_]](
+    semanticdbEnabled := true,
+    semanticdbVersion := scalafixSemanticdb.revision,
+    scalafixScalaBinaryVersion := CrossVersion.binaryScalaVersion(scalaVersion.value),
+    tlScalafixDependencies := Seq(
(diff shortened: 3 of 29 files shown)


core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
```diff
@@ -73,8 +75,8 @@ object TypelevelPlugin extends AutoPlugin {
     },
     githubWorkflowBuild := {
       WorkflowStep.Sbt(
-        List("headerCheckAll", "scalafmtCheckAll", "project /", "scalafmtSbtCheck"),
-        name = Some("Check headers and formatting"),
+        List("headerCheckAll", "scalafmtCheckAll", "scalafixAll --check", "project /", "scalafmtSbtCheck"),
+        name = Some("Check headers, migrations and formatting"),
```
armanbilge, line 79: So actually after the discussion in https://github.com/typelevel/sbt-typelevel/issues/247#issuecomment-1092004966 I had an idea to do these differently.

Basically, I think the `TypelevelCiPlugin` should expose boolean settings for enabling header, fmt, scalafix, and mima checks in CI and then generate the CI steps based on those. Then, builds and/or other plugins can enable those settings (or if we are really clever, they can introspect plugins available on the build and set themselves).

This will also make it easier for an individual build to decide whether to opt-in or opt-out of any of these particular checks.

armanbilge: Oh, another thing to note here, is that `scalafixAll --check` I believe needs a full compile of the project, which is why it's usually called as a later step in the build.

armanbilge:
> if we consider cats, there are currently some bincompat tricks in place that conflict with the simulacrum-scalafix generated code, so it's currently (deliberately) not in sync with its Scalafix migrations.

IIUC it seems the problem is that some of the configured scalafix rules are not actually ones that we want to run/test in CI.
https://github.com/typelevel/cats/blob/3cb763920f177f045e2253be51e52ea2554770ad/.scalafix.conf

It seems to me we may want to have two separate scalafix configs. We do this in http4s for example so we can relax some of the linting rules for test sources.
https://github.com/http4s/http4s/blob/9ac020a92409ce86ad8d67c08ba4dcd31bb6438b/.scalafix.test.conf

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
DavidGregory084: Ah of course, the semanticdb doesn't exist until then anyway! I guess it ought to be a build step then.

scalafix/src/main/scala/org/typelevel/sbt/TypelevelScalafixPlugin.scala
```diff
@@ -0,0 +1,30 @@
...
+  import autoImport._
+
+  override def buildSettings = Seq[Setting[_]](
+    semanticdbEnabled := true,
+    semanticdbVersion := scalafixSemanticdb.revision,
+    scalafixScalaBinaryVersion := CrossVersion.binaryScalaVersion(scalaVersion.value),
+    tlScalafixDependencies := Seq(
+      "com.github.liancheng" %% "organize-imports" % "0.6.0"
+    ),
+    scalafixDependencies ++= tlScalafixDependencies.value
```
armanbilge, line 28: Is there a reason we need our own `tl` setting for this one? It seems reasonable to just append to the existing setting.

DavidGregory084: No reason, that just seemed to be the convention in this plugin e.g. `tlMimaPreviousVersions` 😛 I will remove it

armanbilge: Ah, that's because there is no such equivalent setting in MiMa itself :) see https://github.com/typelevel/sbt-typelevel/pull/176.

.scalafix.conf
```diff
@@ -0,0 +1,7 @@
+rules = [
+    OrganizeImports
+]
+
+OrganizeImports {
+  preset = INTELLIJ_2020_3
+}
```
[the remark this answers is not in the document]
DavidGregory084: I chose this configuration because although VS Code users can `Organize Imports` straight from their editor, there will always be folks using IntelliJ which has its own way of handling import organisation, so this seems like a good choice to minimise friction for contributors.

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
DavidGregory084: I had a quick look at this and this is where I got to:
```diff
diff --git a/ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala b/ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
index 335f99a..53385fa 100644
--- a/ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
+++ b/ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
@@ -20,33 +20,87 @@ import sbt._
 import org.typelevel.sbt.gha.GenerativePlugin
 import org.typelevel.sbt.gha.GitHubActionsPlugin
 import org.typelevel.sbt.gha.GenerativePlugin.autoImport._
+import org.typelevel.sbt.TypelevelScalafixPlugin.autoImport._
 import com.typesafe.tools.mima.plugin.MimaPlugin
 import scala.language.experimental.macros
 
 object TypelevelCiPlugin extends AutoPlugin {
 
-  override def requires = GitHubActionsPlugin && GenerativePlugin && MimaPlugin
+  override def requires =
+    GitHubActionsPlugin && GenerativePlugin && MimaPlugin && TypelevelScalafixPlugin
+
   override def trigger = allRequirements
 
   object autoImport {
     def tlCrossRootProject: CrossRootProject = macro CrossRootProjectMacros.crossRootProjectImpl
+
+    val tlCheckHeaders =
+      settingKey[Boolean]("Check for presence of copyright headers in CI (default: true)")
+    val tlCheckFormatting = settingKey[Boolean](
+      "Check code is formatted according to the project's Scalafmt config in CI (default: true)")
+    val tlCheckMigrations = settingKey[Boolean](
+      "Check configured Scalafix migrations have been applied in CI (default: true)")
   }
 
+  import autoImport._
+
   override def buildSettings = Seq(
+    tlCheckHeaders := true,
+    tlCheckFormatting := true,
+    tlCheckMigrations := true,
     githubWorkflowPublishTargetBranches := Seq(),
-    githubWorkflowBuild := Seq(
-      WorkflowStep.Sbt(List("test"), name = Some("Test")),
-      WorkflowStep.Sbt(
-        List("mimaReportBinaryIssues"),
-        name = Some("Check binary compatibility"),
-        cond = Some(primaryJavaCond.value)
-      ),
-      WorkflowStep.Sbt(
-        List("doc"),
-        name = Some("Generate API documentation"),
-        cond = Some(primaryJavaCond.value)
+    githubWorkflowBuild := {
+      val checkHeadersCommands =
+        if (tlCheckHeaders.value) List("headerCheckAll") else Nil
+
+      val checkFormattingCommands =
+        if (tlCheckFormatting.value)
+          List("scalafmtCheckAll", "project /", "scalafmtSbtCheck")
+        else
+          Nil
+
+      val checkMigrationsStep =
+        if (tlCheckMigrations.value)
+          Seq(
+            WorkflowStep.Sbt(
+              List("scalafixAll --check"),
+              name = Some("Check Scalafix migrations"),
+              cond = Some(primaryJavaCond.value)
+            )
+          )
+        else Nil
+
+      val preBuildSteps = Seq(
+        WorkflowStep.Sbt(
+          checkHeadersCommands ++ checkFormattingCommands,
+          name = Some("Check headers and formatting"),
+          cond = Some(primaryJavaCond.value)
+        )
       )
-    ),
+
+      val buildStep = Seq(
+        WorkflowStep.Sbt(List("test"), name = Some("Test"))
+      )
+
+      val postBuildSteps = checkMigrationsStep ++ Seq(
+        WorkflowStep.Sbt(
+          List("mimaReportBinaryIssues"),
+          name = Some("Check binary compatibility"),
+          cond = Some(primaryJavaCond.value)
+        ),
+        WorkflowStep.Sbt(
+          List("doc"),
+          name = Some("Generate API documentation"),
+          cond = Some(primaryJavaCond.value)
+        )
+      )
+
+      Seq(
+        preBuildSteps,
+        buildStep,
+        postBuildSteps
+      ).flatten
+    },
     githubWorkflowJavaVersions := Seq(JavaSpec.temurin("8"))
   )
 
@@ -54,5 +108,4 @@ object TypelevelCiPlugin extends AutoPlugin {
     val java = githubWorkflowJavaVersions.value.head
     s"matrix.java == '${java.render}'"
   }
-
 }
```
I need to take a break now though and I seem to have lost the artifact upload steps, so I think I will wait to see what your changes look like instead 😆

.scalafix.conf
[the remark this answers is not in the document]
rossabaker: I might otherwise agree, but the IntelliJ preset's [correctness issue](https://github.com/liancheng/scalafix-organize-imports#210-preset) concerns me.  And since most projects I maintain end up with maintenance branches, I appreciate the care the defaults put into diff reduction.

I won't die on this hill, though.

Issue #177, opened 2022-02-20
Don't run CI on non-primary branches


armanbilge:
Basically everything goes through PRs nowadays, so it effectively causes already large matrices to double in size. So I think this might be a better default and a project can always manually enable CI for all branches if they want.

Thanks to @bblfish for reminding me about this.

armanbilge:
FYI a manual workaround for this is to add to your build:
```scala
ThisBuild / githubWorkflowTargetBranches := Seq("main")
```

rossabaker:
Once in a rare while, it's insightful to see that there is a "semantic merge conflict", where main and the branch pass but the merge commit does not.  But it's probably not worth doubling CI for.  Most conflicts manifest as git conflicts.

armanbilge:
Yes, I've found that occasionally helpful as well.

Actually, there is another use for this: when you working off a branch in your own repo, it appears to double CI. But, when you are contributing from your own fork, it enables CI to run on the branch in your fork, possibly while you are waiting for CI approval. So that's something to think about, although I still think the answer is to relax to restrictions on contributors.

armanbilge:
This just became a serious issue now that typelevel-steward does updates as branches in our repos, rather than in its own fork. Maybe time to think about that 0.5.x release 😓

armanbilge:
Ok, good news on the steward issue: GHA supports a `branches-ignore` option, which we can set to `update/*`. This seems innocuous enough to slip into a patch, and I can put off 0.5.0 again!

armanbilge:
Inspired by the solution for the Steward fiasco, I've proposed a fix for this issue in https://github.com/typelevel/sbt-typelevel/pull/292. Basically, it disables CI on branches prefixed with `pr/`.

This should be a sufficient escape-hatch for those who want this. Meanwhile, most branches will continue to get CI, which enables forks to run CI prior to creating a PR (I use this all the time) and also for non-primary branches to easily publish snapshots (I also do this all the time).

Also, it's sufficiently backwards-compatible that it will land in 0.4.12 in the near future.

PR #291, opened 2022-06-05
Don't run CI on `update/*` branches

armanbilge:
See https://github.com/typelevel/sbt-typelevel/issues/177#issuecomment-1146580184 for context.

Strategically named this branch `update/ignore-me` to see if it works right.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,9 +9,9 @@ name: Continuous Integration
 
 on:
   pull_request:
-    branches: ['**']
+    branches: ['**', '!update/**']
   push:
-    branches: ['**']
+    branches: ['**', '!update/**']
     tags: [v*]
 
 env:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -9,9 +9,9 @@ name: Continuous Integration
 
 on:
   pull_request:
-    branches: ['**']
+    branches: ['**', '!update/**']
```
armanbilge, line 12: This took me aback at first, because I thought it meant it would not run CI for PRs from branches named `update/**`. But, no, it won't run CI for PRs targeting branches named `update/**`.

danicheg: Thanks for clarifying this. Can't imagine how many times people were confused with this. Maybe it's just an attempt to get more traffic to docs? 😆

PR #294, opened 2022-06-08
Add typelevel-scalafix to the sbt-typelevel-scalafix dependencies

### LINKED ISSUE #293: Ship typelevel-scalafix rules in sbt-typelevel-scalafix
Super excited about this :)

DavidGregory084:
Resolves #293

### DIFF
--- a/scalafix/src/main/scala/org/typelevel/sbt/TypelevelScalafixPlugin.scala
+++ b/scalafix/src/main/scala/org/typelevel/sbt/TypelevelScalafixPlugin.scala
@@ -28,9 +28,21 @@ object TypelevelScalafixPlugin extends AutoPlugin {
 
   override def trigger = allRequirements
 
+  object autoImport {
+    val tlTypelevelScalafixVersion = settingKey[String](
+      "The version of typelevel-scalafix to add to the scalafix dependency classpath."
+    )
+  }
+
+  import autoImport._
+
   override def buildSettings = Seq[Setting[_]](
     semanticdbEnabled := true,
     semanticdbVersion := scalafixSemanticdb.revision,
-    SettingKey[Boolean]("tlCiScalafixCheck") := true
+    SettingKey[Boolean]("tlCiScalafixCheck") := true,
+    tlTypelevelScalafixVersion := "0.1.1",
+    scalafixDependencies ++= Seq(
+      "org.typelevel" %% "typelevel-scalafix" % tlTypelevelScalafixVersion.value
+    )
   )
 }
--- a/scalafix/src/main/scala/org/typelevel/sbt/TypelevelScalafixPlugin.scala
+++ b/scalafix/src/main/scala/org/typelevel/sbt/TypelevelScalafixPlugin.scala
@@ -29,8 +29,8 @@ object TypelevelScalafixPlugin extends AutoPlugin {
   override def trigger = allRequirements
 
   object autoImport {
-    val tlTypelevelScalafixVersion = settingKey[String](
-      "The version of typelevel-scalafix to add to the scalafix dependency classpath."
+    val tlTypelevelScalafixVersion = settingKey[Option[String]](
+      "The version of typelevel-scalafix to add to the scalafix dependency classpath, or None to omit the dependency entirely."
     )
   }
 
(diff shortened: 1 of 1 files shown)


armanbilge (commented):
🚀 but this does mean we have to be serious about typelevel-scalafix backwards compatibility from this point :)

scalafix/src/main/scala/org/typelevel/sbt/TypelevelScalafixPlugin.scala
```diff
@@ -28,9 +28,21 @@ object TypelevelScalafixPlugin extends AutoPlugin {
 
   override def trigger = allRequirements
 
+  object autoImport {
+    val tlTypelevelScalafixVersion = settingKey[String](
+      "The version of typelevel-scalafix to add to the scalafix dependency classpath."
+    )
```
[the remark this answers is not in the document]
DavidGregory084: Yeah that makes sense to me!

scalafix/src/main/scala/org/typelevel/sbt/TypelevelScalafixPlugin.scala
```diff
@@ -28,9 +28,21 @@ object TypelevelScalafixPlugin extends AutoPlugin {
...
+
   override def buildSettings = Seq[Setting[_]](
     semanticdbEnabled := true,
     semanticdbVersion := scalafixSemanticdb.revision,
-    SettingKey[Boolean]("tlCiScalafixCheck") := true
+    SettingKey[Boolean]("tlCiScalafixCheck") := true,
+    tlTypelevelScalafixVersion := "0.1.1",
+    scalafixDependencies ++= Seq(
+      "org.typelevel" %% "typelevel-scalafix" % tlTypelevelScalafixVersion.value
+    )
```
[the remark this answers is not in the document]
armanbilge: I meant since it is `typelevel-scalafix` and not `typelevel-scalafix-cats`.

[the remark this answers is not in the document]
DavidGregory084: Yes of course! Forgot that there's no aggregate module yet

scalafix/src/main/scala/org/typelevel/sbt/TypelevelScalafixPlugin.scala
```diff
@@ -28,9 +28,26 @@ object TypelevelScalafixPlugin extends AutoPlugin {
 
   override def trigger = allRequirements
 
+  object autoImport {
+    val tlTypelevelScalafixModules = settingKey[Seq[String]](
+      "The typelevel-scalafix modules to add to the scalafix dependency classpath, or Seq.empty to omit them entirely. See available modules at https://github.com/typelevel/typelevel-scalafix."
+    )
```
DavidGregory084, line 34: How about something like this instead @armanbilge - so that users can pick modules rather than making the version optional?

armanbilge: Hmm. I think the fact that rules have to be activated individually anyway suggests a single aggregate module would work fine here. Otherwise maintaining a rule list and a module list starts to be a bit much I think. (Also we'd have to keep the default in sbt-typelevel up-to-date).

DavidGregory084: Cool, let's add an aggregate module in typelevel-scalafix then

armanbilge: >  Or is this just about typelevel-scalafix's rules

It is, but we're going back on this. I think we should just publish all the rules under one aggregate module, because you still have to activate them individually in your scalafix config.

DavidGregory084: I've reverted this change as we now depend on the aggregate module instead :D

scalafix/src/main/scala/org/typelevel/sbt/TypelevelScalafixPlugin.scala
```diff
@@ -28,9 +28,22 @@ object TypelevelScalafixPlugin extends AutoPlugin {
...
+  }
+
+  import autoImport._
+
   override def buildSettings = Seq[Setting[_]](
     semanticdbEnabled := true,
     semanticdbVersion := scalafixSemanticdb.revision,
-    SettingKey[Boolean]("tlCiScalafixCheck") := true
+    SettingKey[Boolean]("tlCiScalafixCheck") := true,
+    tlTypelevelScalafixVersion := Some("0.1.2"),
```
armanbilge, line 43: Somehow I doubt if we'll get steward updates for this. We'll have to remember to bump I guess.

Issue #268, opened 2022-05-02
Smarter caching in CI


armanbilge:
~~Two~~ Three things to consider:
- cache all deps, not just sbt deps
- be robust to different ci jobs that may download different deps (e.g., jvm vs js vs scalafix vs site)
- don't get invalidated by simple changes to`build.sbt` etc.

armanbilge:
One idea to fix might be to determine the cache key not just from the build configuration, but also from the matrix. Effectively each job in the matrix would get its own cache.

GH actions supports caches of up to 10 GB before it starts evicting old caches. If each job gets its own cache this could add up real quick, at which point it would be useless.

DavidGregory084:
> GH actions supports caches of up to 10 GB before it starts evicting old caches. If each job gets its own cache this could add up real quick, at which point it would be useless.

Perhaps this could be mitigated by using [restore keys](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key=), provided the keys were prefixed in a way that allowed the cache to be layered nicely

armanbilge:
That's a good point. I'd say if we can get sbt and plugins out of the cache then it's a win. Library deps are icing on the cake :)

armanbilge:
Seems we may be caching some files we are not supposed to, like `*.lock` and `ivydata-*.properties`. This should be fixed as well.
https://www.scala-sbt.org/1.x/docs/GitHub-Actions-with-sbt.html#Caching

armanbilge:
The setup-java action now supports caching. I think we should just use that and call it a day. 
- https://github.com/typelevel/sbt-typelevel/pull/300

Issue #26, opened 2021-12-30
Add some tests, scripted and non-scripted


armanbilge:
I think there are some good ones in sbt-spiewak we can adapt. Haven't checked the davenverse.

armanbilge:
So I've been lazy about this, but I haven't forgotten about it. I've had an idea brewing:

1. **There's only so much we can test "internally".**
    For example, sbt-github-actions has tests that check the contents of the generated workflow. However, the problem is the workflow can _change_ (and frequently does). So then these tests have to be updated which is annoying and it's not really a functional test anyway i.e. actually _runs_ the workflow.

2. **For better or worse, the real test is when this actually rolls out.**
     We get along okay by virtue of the fact that sbt-typelevel is self-hosted, but obviously this doesn't test everything. Ideally, we would PR repos with the update _before_ we roll out the official release, to check that everything is working.

3. **Which leads me to think, what we actually want are integration tests.** Something like this:
    - We make our own forks of some representative Typelevel repos, that depend on a snapshot of sbt-typelevel.
    - Every time sbt-typelevel releases a snapshot, we run a custom [scala-steward-action](https://github.com/scala-steward-org/scala-steward-action) to PR the snapshot update to these forks.
    - We monitor these PRs to make sure nothing is broken, particularly before releasing a new version of sbt-typelevel.
    
    These forks can obviously be extremely slimmed down: we don't care so much about the project contents, but rather their real-world build configurations and git histories (e.g., to verify the mima previous version detection is working).

    One last bit of trickyness is testing the release process itself, since the best way to do that is actually with a release. It's probably okay to leave that to sbt-typelevel itself, since it would be difficult to release a version with broken releases 😆 . _However_, it would still be good to exercise the artifact upload/download steps on a dummy publication job, since those are tricky on custom build matrices.

    This will take a little bit of infrastructure, but actually doesn't seem too out-of-reach.

PR #300, opened 2022-06-12
Setup java cache

pjfanning:
Relates to https://github.com/djspiewak/sbt-github-actions/issues/104

Sbt plugins are not my area of expertise. If someone could give me some pointers on how to add test coverage that would be great.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,18 +48,7 @@ jobs:
           distribution: jdkfile
           java-version: 8
           jdkFile: ${{ steps.download-java-temurin-8.outputs.jdkFile }}
-
-      - name: Cache sbt
-        uses: actions/cache@v3
-        with:
-          path: |
-            ~/.sbt
-            ~/.ivy2/cache
-            ~/.coursier/cache/v1
-            ~/.cache/coursier/v1
-            ~/AppData/Local/Coursier/Cache/v1
-            ~/Library/Caches/Coursier/v1
-          key: ${{ runner.os }}-sbt-cache-v2-${{ hashFiles('**/*.sbt') }}-${{ hashFiles('project/build.properties') }}
+          cache: sbt
 
       - name: Check that workflows are up to date
         run: sbt 'project ${{ matrix.project }}' '++${{ matrix.scala }}' 'project /' githubWorkflowCheck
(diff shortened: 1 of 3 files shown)


armanbilge:
Hi, thanks for the PR! This is very exciting, it looks like support for sbt was added recently :)
- https://github.com/actions/setup-java/pull/302

Have you been using this feature in your projects? I ask because it seems there is still some WIP fixes for sbt so I am not sure how stable it is yet.
- https://github.com/actions/setup-java/pull/332

Although, I think our custom cache step currently has the same bug 😅 

> If someone could give me some pointers on how to add test coverage that would be great.

Truthfully this project is shamefully lacking in tests. See https://github.com/typelevel/sbt-typelevel/issues/26. Because the plugin is self-hosting I try to dog-food as much as possible and I think that is sufficient in this case.

armanbilge:
One shortcoming of this change is that caching would no longer be supported for anyone who uses a specific GraalVM version, since that doesn't use the setup-java action.
https://github.com/typelevel/sbt-typelevel/blob/b4961455ad5c0071b5cfc1ca0f6376c88db83d10/github-actions/src/main/scala/org/typelevel/sbt/gha/WorkflowStep.scala#L39-L45

armanbilge:
Sorry, one more question about this: is there any way to customize the cache key? This has been on the todo because we've been having problems with reliability of caching. See:
- https://github.com/typelevel/sbt-typelevel/issues/268

pjfanning:
@armanbilge Good point about the caching when graalvm is used. The action documentation for setup-graalvm does not mention support for a cache param. So, this PR will need to be changed to re-enable the cache action when only graalvm is used in the build.

I use the setup-java caching in some of my projects but not for sbt since I use this sbt plugin. I'll have a closer look at the setup-java implementation to see if there are more issues.

Customising the cache key may be an enhancement that could be added to setup-java. I might raise an issue in that project.

armanbilge:
Awesome. Thanks for your attention to detail, much appreciated! :)

armanbilge (approved):
Thank you very much for this! After doing some research I feel confident with this change.

I will fix caching for GraalVM in follow-up for https://github.com/typelevel/sbt-typelevel/issues/301.

PR #312, opened 2022-06-20
Re-enable unused warnings for scala 2.12

### LINKED ISSUE #307: Enabled unused warnings on Scala 2.12
Somehow these fell through the cracks. H/t @satorg

satorg:
Closes #307 (hopefully)

### DIFF
--- a/core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
+++ b/core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
@@ -110,10 +110,4 @@ object TypelevelPlugin extends AutoPlugin {
 
   // override for bincompat
   override def projectSettings = immutable.Seq.empty
-
-  private val primaryJavaCond = Def.setting {
-    val java = githubWorkflowJavaVersions.value.head
-    s"matrix.java == '${java.render}'"
-  }
-
 }
--- a/github-actions/src/main/scala/org/typelevel/sbt/gha/GenerativePlugin.scala
+++ b/github-actions/src/main/scala/org/typelevel/sbt/gha/GenerativePlugin.scala
@@ -614,7 +614,6 @@ ${indent(jobs.map(compileJob(_, sbt)).mkString("\n\n"), 1)}
             key -> values.take(1) // we only want the primary value
       }
 
-      val keys = "scala" :: additions.keys.toList.sorted
       val oses = githubWorkflowOSes.value.toList
       val scalas = githubWorkflowScalaVersions.value.toList
       val javas = githubWorkflowJavaVersions.value.toList
--- a/mergify/src/main/scala/org/typelevel/sbt/mergify/MergifyAction.scala
+++ b/mergify/src/main/scala/org/typelevel/sbt/mergify/MergifyAction.scala
@@ -18,6 +18,7 @@ package org.typelevel.sbt.mergify
 
 import io.circe.Encoder
 import io.circe.syntax._
+import scala.annotation.nowarn
 
 sealed abstract class MergifyAction {
   private[mergify] def name = getClass.getSimpleName.toLowerCase
@@ -57,6 +58,6 @@ object MergifyAction {
       }
   }
 
+  @nowarn("cat=unused")
   private[this] object Dummy extends MergifyAction
-
 }
--- a/mergify/src/main/scala/org/typelevel/sbt/mergify/MergifyAction.scala
+++ b/mergify/src/main/scala/org/typelevel/sbt/mergify/MergifyAction.scala
@@ -18,6 +18,7 @@ package org.typelevel.sbt.mergify
 
 import io.circe.Encoder
 import io.circe.syntax._
+
 import scala.annotation.nowarn
 
 sealed abstract class MergifyAction {
--- a/settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
+++ b/settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
@@ -21,15 +21,14 @@ import com.github.sbt.git.SbtGit.git
 import org.typelevel.sbt.kernel.GitHelper
 import org.typelevel.sbt.kernel.V
 import sbt._
-import sbt._
 import sbtcrossproject.CrossPlugin.autoImport._
 import sbtcrossproject.CrossType
 
 import java.io.File
 import java.lang.management.ManagementFactory
+import scala.annotation.nowarn
 import scala.util.Try
 
-import Keys._
 import Keys._
 
 object TypelevelSettingsPlugin extends AutoPlugin {
@@ -317,4 +316,7 @@ object TypelevelSettingsPlugin extends AutoPlugin {
       oldSchool ++ newSchool ++ old
     }
   }
+
+  @nowarn("cat=unused")
+  private[this] def unused(): Unit = ()
 }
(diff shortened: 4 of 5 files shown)


settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
```diff
@@ -86,23 +86,20 @@ object TypelevelSettingsPlugin extends AutoPlugin {
       val warnings211 =
         Seq("-Ywarn-numeric-widen") // In 2.10 this produces a some strange spurious error
 
-      val warnings212 = Seq("-Xlint:-unused,_")
+      val warnings212 = Seq.empty[String]
 
-      val removed213 = Set("-Xlint:-unused,_", "-Xlint")
+      val removed213 = Set(
+        "-Xlint",
```
satorg, line 92: Why do we remove `-Xlint` for 2.13?

DavidGregory084: I don't believe that `-Xlint` on its own exists in 2.13 - I think it was replaced by the multi-argument one `-Xlint:_`, but I could be misremembering

satorg: Hmm.. I tried briefly both `-Xlint` and `-Xlint:_` for both Scala 2.12 and 2.13. And seems all these combinations work in a similar way:
```
$ scala-cli compile -j 11 -S 2.12 '-Xlint' test1.scala 
[warn] ./test1.scala:2:31: parameter value ord in method hello is never used
[warn]   def hello[A](a: A)(implicit ord: Ordering[A]): Unit = {
[warn]                               ^
$ scala-cli compile -j 11 -S 2.12 '-Xlint:_' test1.scala 
[warn] ./test1.scala:2:31: parameter value ord in method hello is never used
[warn]   def hello[A](a: A)(implicit ord: Ordering[A]): Unit = {
[warn]                               ^
$ scala-cli compile -j 11 -S 2.13 '-Xlint' test1.scala 
[warn] ./test1.scala:2:31: parameter value ord in method hello is never used
[warn]   def hello[A](a: A)(implicit ord: Ordering[A]): Unit = {
[warn]                               ^^^^^^^^^^^^^^^^
$ scala-cli compile -j 11 -S 2.13 '-Xlint:_' test1.scala 
[warn] ./test1.scala:2:31: parameter value ord in method hello is never used
[warn]   def hello[A](a: A)(implicit ord: Ordering[A]): Unit = {
[warn]                               ^^^^^^^^^^^^^^^^
~
```
Although it is hard to say empirically whether they are fully interchangeable or not.

satorg: Anyway, removing `-Xlint` there is no `-Xlint:_` added for 2.13.

github-actions/src/main/scala/org/typelevel/sbt/gha/GenerativePlugin.scala
```diff
@@ -614,7 +614,6 @@ ${indent(jobs.map(compileJob(_, sbt)).mkString("\n\n"), 1)}
             key -> values.take(1) // we only want the primary value
       }
 
-      val keys = "scala" :: additions.keys.toList.sorted
```
satorg, line 617: unused, so began failing after re-enabling the warnings.

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
```diff
@@ -110,10 +110,4 @@ object TypelevelPlugin extends AutoPlugin {
 
   // override for bincompat
   override def projectSettings = immutable.Seq.empty
-
-  private val primaryJavaCond = Def.setting {
-    val java = githubWorkflowJavaVersions.value.head
-    s"matrix.java == '${java.render}'"
-  }
-
```
satorg, line 118: Unused. Seems it was copied to here:
https://github.com/typelevel/sbt-typelevel/blob/c810fa9f8d3d71c9fc18158537699ac7f62891e1/ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala#L129-L132

satorg:
@armanbilge I think it is pretty much ready, could you take a look please?

Btw, there are no "unused" warnings enabled neither for Scala 3.0 nor 3.1 as well.

armanbilge:
Thanks for all your work on this! Much appreciated.

To make up for the lack of tests 😬 I usually try to dog-food things to confirm they are working. Along those lines maybe we can add something like this inside the `TypelevelSettingsPlugin`:
```scala
@nowarn("cat=unused")
private[this] def unused: Unit = ()
```

Also, as usual I am terrified what will happen when we roll this out to 100+ projects 😅 but assuming that nearly all projects cross-building for 2.12 are also building for 2.13, and there are no weird differences in unused warnings, it should be okay.

> Btw, there are no "unused" warnings enabled neither for Scala 3.0 nor 3.1 as well.

Unfortunately that's because nobody has implemented them in dotty yet 😕

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
armanbilge: Nice!

github-actions/src/main/scala/org/typelevel/sbt/gha/GenerativePlugin.scala
armanbilge: Also nice!

mergify/src/main/scala/org/typelevel/sbt/mergify/MergifyAction.scala
```diff
@@ -57,6 +57,6 @@ object MergifyAction {
       }
   }
 
+  @scala.annotation.nowarn("cat=unused")
   private[this] object Dummy extends MergifyAction
```
armanbilge, line 61: Cool, seems we basically already have the test I'm looking for 😂

satorg:
> Along those lines maybe we can add something like this inside the TypelevelSettingsPlugin:
```
@nowarn("cat=unused")
private[this] def unused: Unit = ()
```
In fact, I actually added `@nowarn("cat=unused")` in a couple of places already, because they began failing without it.
So we can say for sure that these changes are working to some extent. 

Do you think it make sense to add one more such an annotation directly just for testing?

satorg:
> Unfortunately that's because nobody has implemented them in dotty yet 😕

Yes, you're right. Although Scala 3.1 has added support for `-Wunused:all` but seems it does nothing.

armanbilge:
> Do you think it make sense to add one more such an annotation directly just for testing?

Sorry, my comments were stuck in pending 😆 yes, I did notice those :) I think a deliberate one in the settings plugin would be good.

armanbilge (approved):
👍 in principle, I don't really understand scalac options 😂  getting @DavidGregory084's sign-off would be great.

I might let this live in a snapshot for a bit, which should be enough to unblock the Cats work and will give me a chance to make sure this doesn't cause any surprises in e.g. http4s.

satorg:
Hmm.. I'm struggling to understand, why has the workflow failed?

armanbilge:
Try running `scalafixAll` ?

satorg:
I did... It succeeds if I run it locally.

armanbilge:
It succeeds, but then makes some changes to commit? :)

armanbilge:
Wooops, I just realized this was targeted at main. If we want to get this into the next 0.4.x release it should be targeted at `series/0.4`.

satorg:
Hmm... Seems that somehow I missed your last comment @armanbilge, sorry. Does it still make sense to backport these changes to `series/0.4`?

armanbilge:
No problem. I was always on the fence about this, since 0.5.x is pretty close (we'll need to bump for sbt 1.7.0 and laika 0.19.0) and I wasn't sure how this will affect some builds in case 2.12 has different warnings for some reason.

That is to say, I think you needed this for you Cats PR, so we can put you on a snapshot while you work on it and then something more stable when its ready to merge.

PR #313, opened 2022-06-21
Add FAQ about ScalaJS version

valencik:
Adds a FAQ about the ScalaJS version management
See https://github.com/typelevel/otel4s/pull/18

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -96,13 +96,13 @@ ThisBuild / tlSonatypeUseLegacyHost := false
 
 Check out the [**sbt-typelevel-site**](site.md) plugin.
 
-## How do I manage my ScalaJS version?
+## How do I manage my Scala.js version?
 
-We recommend explicitly setting your ScalaJS version in `project/plugins.sbt`.
+We recommend explicitly setting your Scala.js version in `project/plugins.sbt`.
 
-```
+```scala
 addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.10.0")
 ```
 
-**sbt-typelevel** ships with a conservative ScalaJS version to enable certain settings.
-If one of your dependencies requires a newer ScalaJS version you may experience failures in the `scalaJSLink` CI step.
+**sbt-typelevel** ships with a conservative Scala.js version to enable certain settings.
+If one of your dependencies requires a newer Scala.js version you may experience failures in the `scalaJSLink` CI step.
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -95,3 +95,14 @@ ThisBuild / tlSonatypeUseLegacyHost := false
 ## How do I publish a website like this one?
 
 Check out the [**sbt-typelevel-site**](site.md) plugin.
+
+## How do I manage my ScalaJS version?
+
+We recommend explicitly setting your ScalaJS version in `project/plugins.sbt`.
+
+```
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.10.0")
```
armanbilge, line 104: We should make this an `@SJS_VERSION@` injected variable. You can access the `scalaJSVersion` key in the build.

armanbilge: And here's an example how to inject it.
https://github.com/http4s/http4s-dom/blob/17db4bf4411c62667e4f93f5e767ba61b3519bb4/build.sbt#L219-L222

armanbilge: Oh wait, but the sbt-typelevel one will always be old. Hum. I guess this is fine.

valencik: Just noting that adding
```
"SJS_VERSION" -> scalaJSVersion,
```
to https://github.com/typelevel/sbt-typelevel/blob/c810fa9f8d3d71c9fc18158537699ac7f62891e1/site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala#L114
Did not work.

But I agree that sbt-typelevel's would basically always be out of date and that might be confusing?
I wonder if we could easily pull the latest version via coursier or something.

Perhaps not worth the effort.

armanbilge: Right, it can't go in the `TypelevelSitePlugin` (which doesn't depend on the `ScalaJSPlugin)` because that would roll it out to everyone's builds :) rather, this would be something specific to the sbt-typelevel `build.sbt`.

> I wonder if we could easily pull the latest version via coursier or something.

Haha temping. I might be sniped on this. I'll merge as is and then maybe give this a try.

armanbilge:
Thank you, much appreciated! This has definitely become an FAQ :)

valencik:
> Oh woops, this merged into main 🤔

Shoot, I neglected to switch the base after you mentioned I should target 0.4

Can I just cherry-pick this onto 0.4?
Have I now made things rough for you when 0.5 releases...

armanbilge:
No worries, cherry-pick is fine. Eh, this will merge cleanly into 0.5 and you'll get credited all over again 😜

Issue #100, opened 2022-01-25
Versioned documentation


rossabaker:
It would be nice to have a turnkey solution for projects that wish to support versioned documentation. 

### Table stakes

* [ ] Each maintained branch gets its own set of docs
* [ ] Can publish from maintenance branches without breaking new things
* [ ] Can publish from main without breaking old things
 
### Nice to have

* [ ] A new version updates the version selector in old versions
* [ ] EOL versions can easily point to a maintained version
* [ ] Version selector jumps to corresponding page, or a fallback if not relevant. (Laika does this.)
* [ ] If a PR will break the site generator, it fails.  (A risk if mdoc compilation is separate from site generation.)
 
### Extra credit

* [ ] `/stable/*` redirects to the latest stable version, like Read The Docs.  This was possible on Netlify, which is an administrative headache.  This is impossible on gh-pages.
* [ ] Works across repos for polyrepo projects (subdomains are a viable alternative) 
* [ ] EOL versions are not indexed by Google.  (Maybe.  This would improve SEO _within_ the site, but could hurt it _to_ the site.)

### http4s

The `/docs` directory on live branches publishes to a subdirectory of the gh-pages branch.  The `/website` directory is published off main and contains the project information that pertains to all branches.

Because each branch publishes independently, there is no way to update the theme across the board, nor a way to automatically update the version selector on old branches when release statuses change.  Care must be taken to not touch anything outside that subdirectory. 

### cats-effect

I don't know how it works.  Submodules and stuff.  Generates crufty PRs, but the end user experience is best-in-Typelevel.

### Monix

Has also been doing it for years in an artisanal fashion.

### Other Scala projects that do a nice job and are worth studying

* Akka
* Play

### Other tooling

Laika has multiple version support, and sbt-typelevel already integrates Laika

## Other ideas

All of these are based on the idea that the site generator runs once, instead of independent sites per version and for the common area.  This raises a basic tradeoff:

* If the site generator runs in the CI of a branch, we have a race condition
* If the site generator is triggered by a workflow after a branch is merged, that branch can break the site

### Publish mdocs per branch, separate job generates the site

Requires careful git exclusions.  Is the most economical with CI time.

### mdoc all the branches and generate the site in one job

Awfully slow for large projects.

### Parallelize building all branches, upload artifacts, generate in separate job

It's a heavy build, but it avoids git exclusions

eed3si9n:
###  `/stable/*` redirects to the latest stable version

[...]

PR #317, opened 2022-06-22
POC: publish the website to sonatype/maven 

armanbilge:
Another strategy for https://github.com/typelevel/sbt-typelevel/issues/100. /cc @rossabaker

The idea is simple, and inspired by the success we've had publishing unidocs. In this case, we package up the Laika-generated website as a Javadoc artifact and publish it. Then, stable versions can be browsed at javadoc.io (with their handy version-picker!) and snapshots via Sonatype's own hosting infrastructure, like so:

https://s01.oss.sonatype.org/service/local/repositories/snapshots/archive/org/typelevel/sbt-typelevel-website_2.12/0.4.12-20-fb939ca-SNAPSHOT/sbt-typelevel-website_2.12-0.4.12-20-fb939ca-SNAPSHOT-javadoc.jar/!/index.html

The motivation is that setting up and maintaining versioned documentation is hard, even with Laika's integrated support for this. It's a lot of overhead for smaller projects that probably have only one active version at a time, but want to keep old documentation available as well.

Now, we have a stable destination to host these old versions of the site. Meanwhile, the current, live version of the site can be hosted on GH pages.

Thoughts?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,15 +89,15 @@ jobs:
         run: sbt 'project ${{ matrix.project }}' '++${{ matrix.scala }}' doc
 
       - name: Make target directories
-        if: github.event_name != 'pull_request' && (startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/series/0.4')
+        if: github.event_name != 'pull_request' && (startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/series/0.4' || github.ref == 'refs/heads/feature/site-artifact')
         run: mkdir -p github/target github-actions/target kernel/target versioning/target ci-release/target scalafix/target target .js/target mdocs/target site/target ci-signing/target mergify/target unidoc/target mima/target .jvm/target .native/target no-publish/target sonatype/target ci/target sonatype-ci-release/target core/target settings/target project/target
 
       - name: Compress target directories
-        if: github.event_name != 'pull_request' && (startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/series/0.4')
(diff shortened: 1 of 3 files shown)


armanbilge:
Thanks! Yes, all credit to Laika again for already doing the right thing with `mappings` :) Note that there is still some work to be done to figure out how to fit this in the CI pipeline, currently the docs are re-built during the publishing step which is sub-optimal.

Good question, let me actually go through Ross's entire checklist:

**Table Stakes**
* [x] _Each maintained branch gets its own set of docs_
    Kind of. In this case, literally every commit has its own docs.
* [x] _Can publish from maintenance branches without breaking new things_
* [x] _Can publish from main without breaking old things_

**Nice to have**

* [x] _A new version updates the version selector in old versions_
    Javadoc.io does this for us :)
* [ ] _EOL versions can easily point to a maintained version_
    No. We cannot change docs once they are published; they are permanent.
* [x] _Version selector jumps to corresponding page, or a fallback if not relevant. (Laika does this.)_
    Yes, javadoc.io also does this :)
* [x] _If a PR will break the site generator, it fails.  (A risk if mdoc compilation is separate from site generation.)_

**Extra credit**

* [ ] _`/stable/*` redirects to the latest stable version, like Read The Docs.  This was possible on Netlify, which is an administrative headache.  This is impossible on gh-pages._
    javadoc.io does support `latest`, but this will give you a milestone. Note that under this strategy, you can just publish stable docs to GH pages and point to javadoc.io to browse other versions.
* [x] _Works across repos for polyrepo projects (subdomains are a viable alternative)_
    TBH not sure what Ross wants here :) but this basically works everywhere.
* [ ] _EOL versions are not indexed by Google.  (Maybe.  This would improve SEO _within_ the site, but could hurt it _to_ the site.)_
    Seems like javadoc.io supports indexing with helpful sitemaps.
    https://javadoc.io/robots.txt

armanbilge:
Gah, I'm not happy with anything after a953d7a162e885f9c06a89d943295f3e09380651. Everything got much more entangled and complicated, hmm.

PR #320, opened 2022-06-30
Add request_reviews and update Mergify actions

bpholt:
Most of our OSS projects use the `request_reviews` and `update` Mergify actions, which aren't supported yet. e.g. https://github.com/Dwolla/fs2-pgp/blob/main/.mergify.yml

We have also typically used merge queues instead of merging directly, but I'm not sure how important that is. I remember we switched from "strict" merges when [Mergify deprecated them](https://blog.mergify.com/strict-mode-deprecation/), but maybe that was overkill for our needs. I'll play with it and see; we can always add that in a future PR.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -8,7 +8,7 @@
 pull_request_rules:
 - name: merge scala-steward's PRs
   conditions:
-  - author=scala-steward
+  - author=typelevel-steward[bot]
   - or:
     - body~=labels:.*early-semver-patch
     - body~=labels:.*early-semver-minor
@@ -168,3 +168,10 @@ pull_request_rules:
       add:
       - versioning
       remove: []
+- name: assign scala-steward's PRs for review
+  conditions:
+  - author=typelevel-steward[bot]
+  actions:
+    request_reviews:
+      users:
+      - armanbilge
--- a/build.sbt
+++ b/build.sbt
@@ -20,7 +20,7 @@ ThisBuild / mergifyPrRules += MergifyPrRule(
   "assign scala-steward's PRs for review",
   List(MergifyCondition.Custom("author=typelevel-steward[bot]")),
   List(
-    MergifyAction.RequestReviews("armanbilge")
+    MergifyAction.RequestReviews.fromUsers("armanbilge")
   )
 )
 
--- a/build.sbt
+++ b/build.sbt
@@ -11,9 +11,18 @@ ThisBuild / developers := List(
   tlGitHubDev("djspiewak", "Daniel Spiewak")
 )
 
-ThisBuild / mergifyStewardConfig ~= { _.map(_.copy(mergeMinors = true)) }
+ThisBuild / mergifyStewardConfig ~= {
+  _.map(_.copy(mergeMinors = true, author = "typelevel-steward[bot]"))
+}
 ThisBuild / mergifySuccessConditions += MergifyCondition.Custom("#approved-reviews-by>=1")
 ThisBuild / mergifyLabelPaths += { "docs" -> file("docs") }
+ThisBuild / mergifyPrRules += MergifyPrRule(
+  "assign scala-steward's PRs for review",
+  List(MergifyCondition.Custom("author=typelevel-steward[bot]")),
+  List(
+    MergifyAction.RequestReviews("armanbilge")
+  )
+)
 
 ThisBuild / scalafixDependencies ++= Seq(
   "com.github.liancheng" %% "organize-imports" % "0.6.0"
--- a/mergify/src/main/scala/org/typelevel/sbt/mergify/MergifyAction.scala
+++ b/mergify/src/main/scala/org/typelevel/sbt/mergify/MergifyAction.scala
@@ -16,7 +16,7 @@
 
 package org.typelevel.sbt.mergify
 
-import io.circe.Encoder
+import io.circe._
 import io.circe.syntax._
 
 sealed abstract class MergifyAction {
@@ -29,6 +29,7 @@ object MergifyAction {
     case merge: Merge => merge.asJson
     case label: Label => label.asJson
     case requestReviews: RequestReviews => requestReviews.asJson
+    case Update => Update.asJson
     case _ => sys.error("should not happen")
   }
 
@@ -65,6 +66,12 @@ object MergifyAction {
   object RequestReviews {
     implicit def encoder: Encoder[RequestReviews] =
       Encoder.forProduct1("users")(_.users)
+  }
+
+  final case object Update extends MergifyAction {
+    override private[mergify] def name = "update"
+
+    implicit def encoder: Encoder[Update.type] = Encoder[JsonObject].contramap(_ => JsonObject.empty)
   }
 
   private[this] object Dummy extends MergifyAction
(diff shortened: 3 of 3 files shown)


bpholt:
Not sure if I targeted the right branch for this PR—feel free to retarget if necessary.

armanbilge:
Thanks! You may want to retarget to `series/0.4`, `main` is going to be 0.5.0, someday ...

bpholt:
I'll rebase and push it up for `series/0.4`

mergify/src/main/scala/org/typelevel/sbt/mergify/MergifyAction.scala
```diff
@@ -57,6 +59,21 @@ object MergifyAction {
...
+  final case class RequestReviews(users: List[String] = Nil) extends MergifyAction {
+    override private[mergify] def name = "request_reviews"
+  }
+
+  object RequestReviews {
+    implicit def encoder: Encoder[RequestReviews] =
+      Encoder.forProduct1("users")(_.users)
+  }
+
+  final case class Update() extends MergifyAction
```
armanbilge, line 71: Would this be better as a `case object`? This is fine, just seems a bit strange :)

bpholt: It might be, yeah. I started off with it as an object and it didn't serialize with the right name, but I figured out how to fix the name on the `RequestReviews`, so it should be doable.

mergify/src/main/scala/org/typelevel/sbt/mergify/MergifyAction.scala
```diff
@@ -57,6 +59,21 @@ object MergifyAction {
       }
   }
 
+  final case class RequestReviews(users: List[String] = Nil) extends MergifyAction {
```
armanbilge, line 62: Can we also add the `teams` and `users_from_teams` keys?. Also frustratingly seems like all of these can be either a `List` or a `Map`. I wonder if we should just go with the `Map` since it subsumes `List` and maybe add some helpful constructors.

https://docs.mergify.com/actions/request_reviews/

I know it's annoying, but fixing these binary-compatibly will be even more annoying :)

bpholt: Good idea, I'll play with this tomorrow or later this weekend.

bpholt: Had some extra time tonight after all. WDYT of what I just pushed up?

(I wondered whether any of these should be case classes if we're worried about bincompat, btw. Should I change the others too?)

armanbilge (commented):
Yeah, `case class`es are a usability vs maintainability trade-off. I think it's okay to use them here, as long as we do our best to include all the members right from the start. Bincompat is not a huge deal for an sbt plugin, but I still like to avoid breaking it if I can.

mergify/src/main/scala/org/typelevel/sbt/mergify/MergifyAction.scala
```diff
@@ -59,16 +61,30 @@ object MergifyAction {
       }
   }
 
-  final case class RequestReviews(users: List[String] = Nil) extends MergifyAction {
+  final class RequestReviews(val users: Either[Seq[String], Map[String, Int]], val randomCount: Option[Int]) extends MergifyAction {
```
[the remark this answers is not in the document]
bpholt: > Also I can't remember how Either renders by default with Circe

I folded them together so that if it's a list it gets rendered without the weights—I was thinking it would be kind of confusing to have weights if they're not specified. I know it's not something users should be editing, but it's still visible and people could be spot-checking it or copy/pasting it into the Mergify config editor for validation or to see what rules would be triggered on different PRs.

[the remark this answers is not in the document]
armanbilge: Ok, fair point. Let's keep it then. How does Circe generate yaml from `JSON`?

Actually, you can add a `RequestReviews` action to the sbt-typelevel mergify config in `build.sbt` so we can dog-food this :)

mergify/src/main/scala/org/typelevel/sbt/mergify/MergifyAction.scala
```diff
@@ -59,16 +61,30 @@ object MergifyAction {
...
+    def apply(user: String, users: String*) =
+      new RequestReviews((user :: users.toList).asLeft, None)
+
+    def apply(weightedUser: (String, Int), weightedUsers: (String, Int)*) =
+      new RequestReviews((weightedUser :: weightedUsers.toList).toMap.asRight, None)
+
+    def apply(users: NonEmptyList[String], randomCount: Int) =
+      new RequestReviews(users.toList.asLeft, Option(randomCount))
+
+    def apply(weightedUsers: NonEmptyMap[String, Int], randomCount: Int) =
```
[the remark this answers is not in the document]
bpholt: Removed from the smart constructors—I agree this would be annoying.

build.sbt
```diff
@@ -14,6 +14,12 @@ ThisBuild / developers := List(
 ThisBuild / mergifyStewardConfig ~= { _.map(_.copy(mergeMinors = true)) }
 ThisBuild / mergifySuccessConditions += MergifyCondition.Custom("#approved-reviews-by>=1")
 ThisBuild / mergifyLabelPaths += { "docs" -> file("docs") }
+ThisBuild / mergifyPrRules += MergifyPrRule("assign scala-steward's PRs for review",
+  List(MergifyCondition.Custom("author=scala-steward")),
```
armanbilge, line 19: While you're here :)

```suggestion
  List(MergifyCondition.Custom("author=typelevel-steward[bot]")),
```

.mergify.yml
```diff
@@ -168,3 +168,14 @@ pull_request_rules:
...
+- name: assign scala-steward's PRs for review
+  conditions:
+  - author=scala-steward
+  actions:
+    request_reviews:
+      users:
+      - armanbilge
+      - rossabaker
+      - ChristopherDavenport
+      - djspiewak
```
armanbilge, line 180: Nice! But let not bother all these folks, just me is fine 😁

mergify/src/main/scala/org/typelevel/sbt/mergify/MergifyAction.scala
```diff
@@ -57,6 +63,102 @@ object MergifyAction {
...
+    override private[mergify] def name = "request_reviews"
+
+    private def copy(
+        users: Option[OptionallyWeighted] = users,
+        teams: Option[OptionallyWeighted] = teams,
+        usersFromTeams: Option[OptionallyWeighted] = usersFromTeams,
+        randomCount: Option[Int] = randomCount): RequestReviews =
+      new RequestReviews(users, teams, usersFromTeams, randomCount) {}
+
+    def andUsers(user: String, users: String*): RequestReviews =
```
bpholt, line 81: Assuming we like this pseudo-builder style, there's a bikeshed opportunity: "and" vs "with"

bpholt: It might also be nice to try to prevent someone from calling `andUsers` if `users` has already been set, but that might be overkill 🙂

armanbilge: Wow, thanks, this looks really really nice! I do like it :)

I'm partial for "with". Since these are collections, "and" makes me thing it's appending (and not replacing). Which actually could be cool too.

armanbilge (approved):
Thanks for all your work on this! It's very nice, and sets a good example of the de-case-classification we'll need to do for feral 😬

PR #322, opened 2022-07-02
Add `doc` to `prePR` command alias

armanbilge:
Also use `session clear` instead of `reload` at the end.


rossabaker (approved):
I don't know what `session clear` is, but the lack of `doc` has bitten me many times over the years.

armanbilge:
`session clear` resets any settings modified in the session (in this case, enabling fatal warnings). It's cheaper than reloading the entire build. H/t https://github.com/typelevel/sbt-typelevel/pull/210#discussion_r827011400

Issue #327, opened 2022-07-10
Using multiple JDKs in CI causes artifact download steps to be duplicated


armanbilge:
As noticed in https://github.com/http4s/http4s-armeria/pull/210#discussion_r917367149.

PR #328, opened 2022-07-10
Only download artifacts for primary OS/JDK

armanbilge:
Fixes https://github.com/typelevel/sbt-typelevel/issues/327.

This also adds a JDK 17 job to CI as part of the replication. I think this is a good addition anyway.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,6 +11,8 @@ ThisBuild / developers := List(
   tlGitHubDev("djspiewak", "Daniel Spiewak")
 )
 
+ThisBuild / githubWorkflowJavaVersions += JavaSpec.temurin("17")
+
 ThisBuild / mergifyStewardConfig ~= {
   _.map(_.copy(mergeMinors = true, author = "typelevel-steward[bot]"))
 }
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -11,6 +11,8 @@ ThisBuild / developers := List(
   tlGitHubDev("djspiewak", "Daniel Spiewak")
 )
 
+ThisBuild / githubWorkflowJavaVersions += JavaSpec.temurin("17")
```
danicheg, line 14: Hmm, you've specified the 17 Java version, but the 8 version is still in GA workflow. Why is so?

github-actions/src/main/scala/org/typelevel/sbt/gha/GenerativePlugin.scala
```diff
@@ -615,9 +615,9 @@ ${indent(jobs.map(compileJob(_, sbt)).mkString("\n\n"), 1)}
       }
 
       val keys = "scala" :: additions.keys.toList.sorted
-      val oses = githubWorkflowOSes.value.toList
+      val oses = githubWorkflowOSes.value.toList.take(1)
```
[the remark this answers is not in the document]
armanbilge: Well, then it would need to be `List(githubWorkflowOSes.value.head)` :)

build.sbt
armanbilge: It's a `+=` so it tests on both 8 (default) and 17. I don't want to stop testing on 8.

danicheg: ah, I see 👍🏻

github-actions/src/main/scala/org/typelevel/sbt/gha/GenerativePlugin.scala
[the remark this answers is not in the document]
danicheg: uh-oh, i was wrong here. you're right!

PR #333, opened 2022-07-15
Expose versioning info in version plugin

### LINKED ISSUE #318: Easy access to `currentRelease` and `currentPreRelease`
Hi!
First of all thanks for the great plugin, it simplifies the project setup a lot!

Lately, I've been setting up https://github.com/ocadotechnology/pass4s where I'm using `sbt-typelevel` for versioning, mima, publishing, and all that good stuff. For the website we have decided to go with docusaurus and mdoc as we already had it in our other project and wanted to be aligned.

For the docs, I wanted the mdoc `VERSION` to point to the latest stable. I've noticed it's already done in `sbt-typelevel-site` in the following lines:
https://github.com/typelevel/sbt-typelevel/blob/947ce2bf07ace1780010512ec7bf9369ab126f67/site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala#L111-L117

Unfortunately, there's no easy way to reuse it. After some trial and error, I've just decided to go with the following workaround 

```scala
def latestStableVersion = {
  import scala.sys.process._
  "git -c versionsort.suffix=- tag --list --sort=-version:refname"
    .!!
    .split("\n")
    .toList
    .map(_.trim)
    .filter(_.startsWith("v"))
    .head
}

/* and in project settings */
mdocVariables := Map(
  "VERSION" -> { if (isSnapshot.value) latestStableVersion else version.value }
),
```

Do you think it's feasible to extract the part about the latest stable release and prerelease as a public part to `sbt-typelevel-version` for this kind of use case?

bplommer:
Resolves #318

### DIFF
--- a/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
+++ b/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
@@ -40,35 +40,6 @@ object TypelevelKernelPlugin extends AutoPlugin {
         }
       })
 
-    private[sbt] lazy val currentRelease: Def.Initialize[Option[String]] = Def.setting {
-      // some tricky logic here ...
-      // if the latest release is a pre-release (e.g., M or RC)
-      // and there are no stable releases it is bincompatible with,
-      // then for all effective purposes it is the current release
-
-      val release = previousReleases.value match {
-        case head :: tail if head.isPrerelease =>
-          tail
-            .filterNot(_.isPrerelease)
-            .find(head.copy(prerelease = None).mustBeBinCompatWith(_))
-            .orElse(Some(head))
-        case releases => releases.headOption
-      }
-
-      release.map(_.toString)
-    }
-
-    // latest tagged release, including pre-releases
-    private[sbt] lazy val currentPreRelease: Def.Initialize[Option[String]] = Def.setting {
-      previousReleases.value.headOption.map(_.toString)
-    }
-
-    private[sbt] lazy val previousReleases: Def.Initialize[List[V]] = Def.setting {
-      val currentVersion = V(version.value).map(_.copy(prerelease = None))
-      GitHelper.previousReleases(fromHead = true, strict = false).filter { v =>
-        currentVersion.forall(v.copy(prerelease = None) <= _)
-      }
-    }
   }
 
   import autoImport._
@@ -112,6 +83,36 @@ object TypelevelKernelPlugin extends AutoPlugin {
         val ver = (LocalRootProject / scalaVersion).value
         (task / tlSkipIrrelevantScalas).value && !cross.contains(ver)
       }
+    }
+  }
+
+  private[sbt] lazy val currentRelease: Def.Initialize[Option[String]] = Def.setting {
+    // some tricky logic here ...
+    // if the latest release is a pre-release (e.g., M or RC)
+    // and there are no stable releases it is bincompatible with,
+    // then for all effective purposes it is the current release
+
+    val release = previousReleases.value match {
+      case head :: tail if head.isPrerelease =>
+        tail
+          .filterNot(_.isPrerelease)
+          .find(head.copy(prerelease = None).mustBeBinCompatWith(_))
+          .orElse(Some(head))
+      case releases => releases.headOption
+    }
+
+    release.map(_.toString)
+  }
+
+  // latest tagged release, including pre-releases
+  private[sbt] lazy val currentPreRelease: Def.Initialize[Option[String]] = Def.setting {
+    previousReleases.value.headOption.map(_.toString)
+  }
+
+  private[this] lazy val previousReleases: Def.Initialize[List[V]] = Def.setting {
+    val currentVersion = V(version.value).map(_.copy(prerelease = None))
+    GitHelper.previousReleases(fromHead = true, strict = false).filter { v =>
+      currentVersion.forall(v.copy(prerelease = None) <= _)
     }
   }
 
--- a/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
+++ b/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
@@ -39,13 +39,13 @@ object TypelevelKernelPlugin extends AutoPlugin {
         }
       })
 
-    private[sbt] lazy val currentReleaseImpl = Def.setting {
+    private[sbt] lazy val currentRelease: Def.Initialize[Option[String]] = Def.setting {
       // some tricky logic here ...
       // if the latest release is a pre-release (e.g., M or RC)
       // and there are no stable releases it is bincompatible with,
       // then for all effective purposes it is the current release
 
-      val release = previousReleasesImpl.value match {
+      val release = previousReleases.value match {
         case head :: tail if head.isPrerelease =>
           tail
             .filterNot(_.isPrerelease)
@@ -58,11 +58,11 @@ object TypelevelKernelPlugin extends AutoPlugin {
     }
 
     // latest tagged release, including pre-releases
-    private[sbt] lazy val currentPreReleaseImpl = Def.setting {
-      previousReleasesImpl.value.headOption.map(_.toString)
+    private[sbt] lazy val currentPreRelease: Def.Initialize[Option[String]] = Def.setting {
+      previousReleases.value.headOption.map(_.toString)
     }
 
-    private[sbt] lazy val previousReleasesImpl = Def.setting {
+    private[sbt] lazy val previousReleases: Def.Initialize[List[V]] = Def.setting {
       val currentVersion = V(version.value).map(_.copy(prerelease = None))
       GitHelper.previousReleases(fromHead = true, strict = false).filter { v =>
         currentVersion.forall(v.copy(prerelease = None) <= _)
--- a/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
+++ b/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
@@ -18,6 +18,7 @@ package org.typelevel.sbt
 
 import sbt._
 import sbt.plugins.JvmPlugin
+
 import Keys._
 import org.typelevel.sbt.kernel.{GitHelper, V}
 
--- a/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
+++ b/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
@@ -16,9 +16,9 @@
 
 package org.typelevel.sbt
 
-import sbt.*
+import sbt._
 import sbt.plugins.JvmPlugin
-import Keys.*
+import Keys._
 import org.typelevel.sbt.kernel.{GitHelper, V}
 
 object TypelevelKernelPlugin extends AutoPlugin {
--- a/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
+++ b/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
@@ -16,11 +16,12 @@
 
 package org.typelevel.sbt
 
+import org.typelevel.sbt.kernel.GitHelper
+import org.typelevel.sbt.kernel.V
 import sbt._
 import sbt.plugins.JvmPlugin
 
 import Keys._
-import org.typelevel.sbt.kernel.{GitHelper, V}
 
 object TypelevelKernelPlugin extends AutoPlugin {
 
--- a/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
+++ b/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
@@ -16,10 +16,10 @@
 
 package org.typelevel.sbt
 
-import sbt._
+import sbt.*
 import sbt.plugins.JvmPlugin
-
-import Keys._
+import Keys.*
+import org.typelevel.sbt.kernel.{GitHelper, V}
 
 object TypelevelKernelPlugin extends AutoPlugin {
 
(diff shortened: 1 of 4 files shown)


kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
```diff
@@ -18,8 +18,8 @@ package org.typelevel.sbt
 
 import sbt._
 import sbt.plugins.JvmPlugin
-
 import Keys._
```
bplommer, line 21:
```suggestion

import Keys._
```

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
```diff
@@ -23,8 +23,8 @@ import org.typelevel.sbt.kernel.V
 import sbt._
 
 import scala.util.Try
-
 import Keys._
```
bplommer, line 26:
```suggestion

import Keys._
```

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
```diff
@@ -37,36 +38,11 @@ object TypelevelVersioningPlugin extends AutoPlugin {
...
-      previousReleases.value.headOption.map(_.toString)
-    }
-
-    lazy val previousReleases = Def.setting {
-      val currentVersion = V(version.value).map(_.copy(prerelease = None))
-      GitHelper.previousReleases(fromHead = true, strict = false).filter { v =>
-        currentVersion.forall(v.copy(prerelease = None) <= _)
-      }
-    }
+    lazy val currentPreRelease = currentPreReleaseImpl
```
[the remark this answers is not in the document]
armanbilge: Also, we should be mindful about naming. `tl`-prefixed to avoid collisions with other plugins.

kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
```diff
@@ -37,6 +38,36 @@ object TypelevelKernelPlugin extends AutoPlugin {
           BasicCommands.addAlias(BasicCommands.removeAlias(s, name), name, contents)
         }
       })
+
+    private[sbt] lazy val currentReleaseImpl = Def.setting {
```
bplommer, line 42: Alternatively we can remove this nasty `Impl` suffix, at the cost of having to avoid full imports at the use site in the versioning plugin

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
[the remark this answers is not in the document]
bplommer: 👍

site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
```diff
@@ -77,7 +75,7 @@ object TypelevelSitePlugin extends AutoPlugin {
   import TypelevelGitHubPlugin._
 
   override def requires =
-    MdocPlugin && LaikaPlugin && TypelevelGitHubPlugin && GenerativePlugin && NoPublishPlugin
+    MdocPlugin && LaikaPlugin && TypelevelGitHubPlugin && TypelevelVersioningPlugin && GenerativePlugin && NoPublishPlugin
```
bplommer, line 78:
```suggestion
    MdocPlugin && LaikaPlugin && TypelevelGitHubPlugin && GenerativePlugin && NoPublishPlugin
```

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
```diff
@@ -37,6 +38,13 @@ object TypelevelVersioningPlugin extends AutoPlugin {
     lazy val tlUntaggedAreSnapshots =
       settingKey[Boolean](
         "If true, an untagged commit is given a snapshot version, e.g. 0.4-00218f9-SNAPSHOT. If false, it is given a release version, e.g. 0.4-00218f9. (default: true)")
+
+    lazy val tlCurrentRelease =
+      settingKey[Option[String]](
+        "The latest stable released version of your project, e.g. 0.2.0, 3.5.1. If applicable, this will be the version currently being released.")
+
+    lazy val tlCurrentPreRelease = settingKey[Option[String]](
+      "The latest pre-release (e.g. milestone, release candidate) of your project. If applicable, this will be the version currently being released.")
```
armanbilge, line 47: Thanks, these are great descriptions! Bikeshed time, is "current release" and "current pre-release" the best terminology for these? Maybe "latest" is more appropriate? And "version" instead of "release"?

I guest that gets us to something like `tlLatestStableVersion` and `tlLatestPreReleaseVersion` maybe 🤔 bit of a mouthful.

bplommer: I have absolutely no opinion on this but happy to change it if you like 😃

armanbilge: Not even a smidgeon of an opinion? 😂 I hate deciding these things. I can commit whatever I decide before I merge.

bplommer: Maybe I have a slight preference for `tlLatestStableRelease` and `tlLatestPreRelease` - explicit but reasonably concise.

armanbilge (commented):
👍 thanks for taking this on! I think it looks great, I will bikeshed the setting names and then merge.

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
bplommer: On the other hand, `currentRelease` isn't actually guaranteed to be a stable release:
> if the latest release is a pre-release (e.g., M or RC)
> and there are no stable releases it is bincompatible with,
> then for all effective purposes it is the current release

armanbilge: Which comment specifically?

The idea is if you are on version `n` and you start prereleasing a binary-breaking version `n+1`, then it has effectively taken over that branch. If you want to keep version `n` alive then it would be on its own branch, where `currentRelease` would continue to be stable versions.

This was designed with http4s in mind: otherwise `currentRelease` on the `main` branch would always be 0.23.x and not the latest 1.0.0 milestone. Meanwhile, `currentRelease` on the `0.23.x` branch will always be a an 0.23.x version.


... but I'm totally open to better ideas here. Part of my hesitation with exposing these was that we need to be pretty solid on what we are exposing :)

bplommer: > Which comment specifically?

The one quoted in https://github.com/typelevel/sbt-typelevel/pull/333#discussion_r922100471

Your explanation makes sense though. So maybe the descriptions should be:

```suggestion
    lazy val tlCurrentRelease =
      settingKey[Option[String]](
        "The latest stable version of your project released from the current branch, e.g. 0.2.0, 3.5.1. If applicable, this will be the version currently being released.")

    lazy val tlCurrentPreRelease = settingKey[Option[String]](
      "The latest version of your project released from the current branch, whether stable or pre-release (e.g. milestone, release candidate). If applicable, this will be the version currently being released.")
```

armanbilge: Yeah, in retrospect I can see calling it a "stable" release when there is no guarantee of that, is really not helpful :)

armanbilge: I think the first one is maybe something like:
> The latest stable version of your project in the `tlBaseVersion` series, with the latest pre-release in this series used as a fallback if there is no stable release yet.

bplommer: It sounds like we need to make up a new word for the concept we're talking about here - `foo` such that https v1.0.0-M34 and v0.23.13 are foo versions but v0.23.14-RC1 would not be a foo version 🤔

bplommer: Maybe "current version" is the best we can do here. And then the description could be:

> The "current release" of your project from the current branch. Usually this will be the latest stable release. However, if your branch contains pre-releases that break binary compatibility with the latest stable release, the branch is considered a pre-release branch and the current release will therefore be the latest pre-release.

It's a bit wordy but I'm not sure how much more concise it's possible to be.

bplommer: Can I leave this with you to commit the final wording and naming? Happy to discuss ideas of course

armanbilge: Yup, I can do that. Thanks for all your help, nice to bounce ideas! Much appreciated.

bplommer: > I think the first one is maybe something like:
> 
> > The latest stable version of your project in the `tlBaseVersion` series, with the latest pre-release in this series used as a fallback if there is no stable release yet.

Ah yeah sounds good, I only just saw this now.

bplommer: > The latest stable version of your project in the `tlBaseVersion` series, with the latest pre-release in this series used as a fallback if there is no stable release yet.

Actually I don't think that's correct? `tlBaseVersion` isn't used in the calculation, and this won't be the behaviour if (say) `tlBaseVersion` is 1.4 but there's a stable 1.3.0 release and a 1.4.0-M1 release, right?

armanbilge: You are right, it's still not 100% correct.

> `tlBaseVersion` isn't used in the calculation

It is, albeit indirectly. `version` is used in the calculation, and `tlBaseVersion` is used to calculate `version`.

> `tlBaseVersion` is 1.4 but there's a stable 1.3.0 release and a 1.4.0-M1 release, right?

You are right about this. It uses the binary series and not the `x.y` series.

kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
```diff
@@ -83,4 +86,34 @@ object TypelevelKernelPlugin extends AutoPlugin {
...
+      case head :: tail if head.isPrerelease =>
+        tail
+          .filterNot(_.isPrerelease)
+          .find(head.copy(prerelease = None).mustBeBinCompatWith(_))
+          .orElse(Some(head))
+      case releases => releases.headOption
+    }
+
+    release.map(_.toString)
+  }
```
armanbilge, line 105: Shoot, still not quite right 😂 I thought this was using `version` in here, but you are right that it's not. Sorry, I was confusing with the change I made in https://github.com/typelevel/sbt-typelevel/pull/282 which does use `version`.

So as it's currently defined, even if you've bumped your base version, if you haven't actually tagged anything with that new base version then it will still use whatever the latest tag is under the old base version. I guess this makes sense. But, now to explain it 🤔 **tl;dr** we do some blackbox heuristics to get you a sensible number, take it or leave it 😛

versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
```diff
@@ -37,6 +38,12 @@ object TypelevelVersioningPlugin extends AutoPlugin {
     lazy val tlUntaggedAreSnapshots =
       settingKey[Boolean](
         "If true, an untagged commit is given a snapshot version, e.g. 0.4-00218f9-SNAPSHOT. If false, it is given a release version, e.g. 0.4-00218f9. (default: true)")
+
+    lazy val tlLatestVersion = settingKey[Option[String]](
+      "The latest tagged version on this branch. Priority is given to the latest stable version, but if you have tagged a binary-breaking prelease version (such as a milestone or release candidate), that will be selected instead. If applicable, this will be the current tagged version.")
+
+    lazy val tlLatestPreReleaseVersion = settingKey[Option[String]](
+      "The latest tagged version on this branch, including milestones and release candidates. If applicable, this will be the current tagged version.")
```
armanbilge, line 46: ok, @bplommer, what do you think? phew 😓

bplommer: I'm happy with that 👍

bplommer: On second thought, what do you think about `tlLatestVersionOnBranch` and `tlLatestPreReleaseVersionOnBranch`? It's a bit wordy, but for me it's really counterintuitive that (for example) these would be 0.22 versions when running on the http4s 0.22.x branch, so they're not really the latest versions.

armanbilge: > but for me it's really counterintuitive that (for example) these would be 0.22 versions when running on the http4s 0.22.x branch

Ok yeah that's super fair. But that is quite a mouthful. What if we used the terminology "previous version" instead of "latest"? Then maybe it would be more clear it's relative to our current context. The problem is that it muddies the fact that this can be the _current_ version it the current commit is tagged.

bplommer: I don't like "previous", for the reason you give - to me that explicitly rules out it being the current one. But I don't have any other suggestions at the moment 🤔

armanbilge: Blah. I think I'm just going to run with the current naming. I think reasonable for the majority of cases (esp. for unbranched projects), and the docs cover it.

> but for me it's really counterintuitive that (for example) these would be 0.22 versions when running on the http4s 0.22.x branch

I guess my question is, what are you doing with `tlLatestVersion` on the 0.22.x branch, where you want it to be an 0.23+ version? My assumption is if you've cut off a series branch, you want to "freeze" it in a sense. So although the setting name doesn't reflect it, I think the actual behavior is intuitive.

bplommer: > I guess my question is, what are you doing with tlLatestVersion on the 0.22.x branch, where you want it to be an 0.23+ version?

I can't think of anything! But I *can* imagine wanting the "latest prerelease" version on the 0.23.x branch to be a 1.0 milestone, so I could populate the website's landing page accordingly.

armanbilge: In that case, shouldn't the landing page be published from `main` instead of `series/0.23`? (This is how we do it in http4s).

armanbilge: Hmm, I guess not necessarily. My proposal in https://github.com/typelevel/sbt-typelevel/pull/317 aims to reduce that complexity, so that you only publish the website from the stable branch. Blah again 😂

bplommer: Blah indeed. Let's just go with it as it is, I think it's good enoigh?

armanbilge: Yolo. We might end up adding a couple more of these as the complaints roll in 😂 thank you so much for your work on this.

Issue #319, opened 2022-06-22
Redirect `your.project.site/api` to javadoc.io API docs


armanbilge:
https://typelevel.org/cats/api/ is easy to remember and an eager suggestion by my browser, but nowadays I really want to go to https://www.javadoc.io/doc/org.typelevel/cats-docs_2.12/latest/

So seems like we should use the meta-refresh trick to get an auto-redirect there and integrate this into the site plugin.

Annoyingly this could be a breaking change if someone was actually putting something at `/api`. But I think Laika would give priority to that. Also I really want this so I might look the other way 😬

PR #335, opened 2022-07-15
Add `/api` path to site that forwards to API url

armanbilge:
Closes https://github.com/typelevel/sbt-typelevel/issues/319.


armanbilge:
Based on my understanding of Laika this is not breaking, userland stuff takes precedent over themes.

Issue #326, opened 2022-07-10
CI job for submitting dependency graph to GitHub


armanbilge:
So this is a really cool idea: https://github.com/ckipp01/mill-github-dependency-graph

It uses new GitHub API described here:
https://github.blog/2022-06-17-creating-comprehensive-dependency-graph-build-time-detection/

Basically it would be a way to get our dependencies registered on this page:
https://github.com/typelevel/sbt-typelevel/network/dependencies

This is beneficial for security/vulnerability alerts. And maybe we can also get one of those cool "used by n" widgets on our repos 😂 

I don't think an sbt version of that plugin should live here, but when it exists, I think we should integrate with it.

armanbilge:
Ah, seems there may be an official announcement soon :)
https://github.com/scalacenter/sbt-dependency-graph-action

PR #340, opened 2022-07-16
Add dependency submission job to CI

armanbilge:
Closes https://github.com/typelevel/sbt-typelevel/issues/326.

This adds a job that submits the projects dependencies to GitHub via
https://github.com/scalacenter/sbt-dependency-submission

Results should go here:
https://github.com/typelevel/sbt-typelevel/network/dependencies

Annoyingly I don't think we'll be able to see any results yet. I've targeted this at the `main` branch since I'm not sure yet how stable the action is, but GitHub only cares about the default branch which is `series/0.4`.

### DIFF
--- a/ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
+++ b/ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
@@ -20,9 +20,12 @@ import com.typesafe.tools.mima.plugin.MimaPlugin
 import org.typelevel.sbt.gha.GenerativePlugin
 import org.typelevel.sbt.gha.GenerativePlugin.autoImport._
 import org.typelevel.sbt.gha.GitHubActionsPlugin
+import org.typelevel.sbt.gha.WorkflowStep
 import sbt._
 
 import scala.language.experimental.macros
+
+import Keys._
 
 object TypelevelCiPlugin extends AutoPlugin {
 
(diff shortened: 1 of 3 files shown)


ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
```diff
@@ -125,7 +132,25 @@ object TypelevelCiPlugin extends AutoPlugin {
...
+          List(
+            WorkflowJob(
+              "dependency-submission",
+              "Submit Dependencies",
+              scalas = List(scalaVersion.value),
+              javas = List(githubWorkflowJavaVersions.value.head),
+              steps = githubWorkflowJobSetup.value.toList :+
+                WorkflowStep
+                  .DependencySubmission
+                  .withCond(Some("github.event_name != 'pull_request'"))
```
armanbilge, line 148: So I'm a bit confused if we want to enable it for PRs or not. This page seems to suggest that if a PR updates dependencies that it can be reviewed, so it really seems like it would need this to run for that PR.

https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-dependency-review

rossabaker (commented):
Oh, cool.  I was just coming to request this feature!!!

ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
rossabaker: The API submissions are not yet integrated into Dependency Review or Organization Insights, but they do trigger Dependabot alerts, even for transitive dependencies (which the POM integration does not!).

armanbilge: I did experiment with dependency review in https://github.com/typelevel/sbt-typelevel/pull/342 but the problem is that PRs from external repos don't have permission to submit the dependencies, so I'm not entirely sure how it would work.

rossabaker: I just introduced a vulnerability in a PR, and nothing flagged it until I merged.  I think push for now, and pull once GitHub better integrates their new toys.

PR #342, opened 2022-07-17
Try out dependency review action

armanbilge:


armanbilge:
Does seem like this works, which answers the question about whether to submit deps for PRs.

PR #349, opened 2022-07-19
Fix console/scalacOptions

satorg:
The `scalacOptions` for `console`/`consoleQuick` commands were broken in #312 and #344
This PR should fix that.


settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
```diff
@@ -71,26 +71,22 @@ object TypelevelSettingsPlugin extends AutoPlugin {
...
       "-feature",
       "-unchecked"),
-    scalacOptions ++= {
-      scalaVersion.value match {
-        case V(V(2, minor, _, _)) if minor < 13 =>
-          Seq("-Yno-adapted-args", "-Ywarn-unused-import")
-        case _ =>
-          Seq.empty
-      }
-    },
```
satorg, line 81: These options are united with the options in the next code block.

PR #356, opened 2022-08-14
Fallback to apiURL if javadocioUrl does not exist

valencik:
Of note, this will yield a URL that does not exist when:
- You run this on a dirty git workspace and thus have a timestamp as part of your snapshot
- You run this on a commit that has not had a snapshot published


armanbilge:
> You run this on a dirty git workspace and thus have a timestamp as part of your snapshot

I'd say this falls into case (2), when it's unpublished. Timestamped snapshots are also publishable :)

armanbilge (approved):
Noice. Thanks!

PR #357, opened 2022-08-17
Adding tests around version handling

valencik:
This PR:
- adds an `munit` test suite to the kernel module
- adds basic tests for version handling and bincompat

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -29,6 +29,8 @@ ThisBuild / scalafixDependencies ++= Seq(
   "com.github.liancheng" %% "organize-imports" % "0.6.0"
 )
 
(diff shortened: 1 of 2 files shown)


armanbilge (approved):
Thanks, this was long long overdue 😅 great job enumerating the cases, I think you've covered all the important permutations.

Also a painful reminder of this very ugly corner of sbt-typelevel 😬 I'd still like to replace it someday e.g. with coursier `Version`.
https://github.com/coursier/coursier/blob/adca57b966a8ad2f4c594abee65ba85c177b26ee/modules/core/shared/src/main/scala/coursier/core/Version.scala

Issue #260, opened 2022-04-17
Disable MiMa for 0.0 series


armanbilge:
Nobody takes 0.0.x seriously, and it's annoying.

_Technically_ a breaking change so better to target at 0.5.0.

armanbilge:
Right! Annoyingly GitHub doesn't close issues if the PR doesn't merge into the default branch.

PR #358, opened 2022-08-22
Do not require binary compat on 0.0.z versions

valencik:
Resolves https://github.com/typelevel/sbt-typelevel/issues/260


armanbilge (approved):
Your new test suite makes this easy to review :)

PR #385, opened 2022-10-08
Do not use deprecated `-target` option on 2.13.9+

armanbilge:
Fixes this error, seen in builds using the `tlJdkRelease` option with Scala 2.13.9.

```
[error] -target is deprecated: Use -release instead to compile against the correct platform API.
```

Relevant PR:
- https://github.com/scala/scala/pull/9982


satorg (approved):
Yes, it definitely makes sense, good call out, thanks 👍

satorg:
I wonder though, should we attempt to address this change in options between v2.13.8 and v2.13.9 in our shiny scala-options version of the future? It's one more thing to think about 🤔

armanbilge:
> I wonder though, should we attempt to address this change in options between v2.13.8 and v2.13.9 in our shiny scala-options version of the future? It's one more thing to think about

Yes, actually this is exactly my hope for scalac-options :)

armanbilge:
Thanks for the review, I really appreciate having a few other eyes around 😃

PR #410, opened 2022-11-07
add a space to the Scala version selector command

bpholt:
Using `sbt '++2.13' test` works, but `sbt '++3' test` does not. However, both `sbt '++ 2.13' test` and `sbt '++ 3' test` work, and the space is included in the [sbt documentation](https://www.scala-sbt.org/1.x/docs/Cross-Build.html#Switching+Scala+version).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,22 +87,22 @@ jobs:
 
       - name: Check headers and formatting
         if: matrix.java == 'temurin@8'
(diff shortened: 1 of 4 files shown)


project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.6.2
+sbt.version=1.7.3
```
armanbilge, line 1: Sorry, I'm trying to keep the 0.4.x series pinned to sbt 1.6 :)

bpholt: Ok, that's fine. I'll drop that commit, it just makes it hard to confirm that the issue is resolved within the project itself.

armanbilge (approved):
Thanks!

PR #416, opened 2022-11-13
Allow configuring `timeout-minutes` for a job and a step

iRevive:
A timeout can be configured for a job and a step.

An example of the step timeout:

```scala
ThisBuild / githubWorkflowBuildPostamble += WorkflowStep.Sbt(
  List("info"),
  timeoutMinutes = Some(10)
)
```

```yml
- timeout-minutes: 10
  run: sbt 'project ${{ matrix.project }}' '++${{ matrix.scala }}' info
```


build.sbt
```diff
@@ -13,6 +13,9 @@ ThisBuild / developers ++= List(
 ThisBuild / githubWorkflowJavaVersions ++=
   Seq(JavaSpec.temurin("17"), JavaSpec(JavaSpec.Distribution.GraalVM("latest"), "17"))
 
+ThisBuild / githubWorkflowBuildTimeoutMinutes := Some(60)
```
iRevive, line 16: These two changes are not necessary. They are here mainly for showcase purposes.

github-actions/src/main/scala/org/typelevel/sbt/gha/GenerativePlugin.scala
```diff
@@ -514,6 +521,7 @@ ${indent(jobs.map(compileJob(_, sbt)).mkString("\n\n"), 1)}
     githubWorkflowBuildMatrixInclusions := Seq(),
     githubWorkflowBuildMatrixExclusions := Seq(),
     githubWorkflowBuildRunsOnExtraLabels := Seq(),
+    githubWorkflowBuildTimeoutMinutes := None,
```
armanbilge, line 524: I wonder if we should make the default `Some(60)`? I think that is sufficient for all Typelevel projects. It would would prevent a hanging build in Cats Effect from taking up 24 runners for 6 hours 😅

iRevive: Makes sense to me

armanbilge (approved):

Issue #381, opened 2022-09-26
Add support for `cancel-in-progress` 


armanbilge:
https://docs.github.com/en/actions/using-jobs/using-concurrency

iRevive:
May we close this one?

armanbilge:
Yep, annoying GitHub 🙄

PR #417, opened 2022-11-13
Allow configuring `concurrency` for a job

iRevive:
Closes https://github.com/typelevel/sbt-typelevel/issues/381.


iRevive:
Unintentionally I proved it works 😅  https://github.com/typelevel/sbt-typelevel/actions/runs/3456431979

build.sbt
```diff
@@ -13,6 +13,10 @@ ThisBuild / developers ++= List(
 ThisBuild / githubWorkflowJavaVersions ++=
   Seq(JavaSpec.temurin("17"), JavaSpec(JavaSpec.Distribution.GraalVM("latest"), "17"))
 
+ThisBuild / githubWorkflowBuildConcurrency := Some(
+  JobConcurrency("ci-${{ matrix.java }}-${{ github.ref }}", cancelInProgress = Some(true))
+)
```
armanbilge, line 18: This is great! Another one that I wonder might make a good default, but I'm not sure.

iRevive: It's relatively easy to block a job by using an incorrect group.
For example, `ci-${{ github.ref }}` does not work with multiple JVMs. And `"ci-${{ matrix.java }}-${{ github.ref }}"` will not work with multiple Scala versions.

On the other hand, if we have complete control of the job matrix, we can generate the correct group:
```scala
val matrix: Map[String, List[String]] = ???
val concurrencyGroup = matrix
  .collect { case (key, values) if values.length > 1 => s"$${{ matrix.$key }}" }
  .mkString("ci-", "-", "")
```

armanbilge: Oh hm I see. We do have access to the full matrix in the plugin, so we should be able to do the right thing.

```scala
case (key, values) if values.length > 1 => s"$${{ matrix.$key }}"
```

Is that condition really necessary, is there any harm to include a key even if there is only one value? Because of the "extra jobs" configuration for the matrix there may be multiple values for that key anyway.

iRevive: I enabled concurrency setting by default.

github-actions/src/main/scala/org/typelevel/sbt/gha/GenerativeKeys.scala
```diff
@@ -51,6 +51,7 @@ trait GenerativeKeys {
     settingKey[Seq[MatrixExclude]]("A list of matrix exclusions (default: [])")
   lazy val githubWorkflowBuildRunsOnExtraLabels = settingKey[Seq[String]](
     "A list of additional labels to append to each run of the matrix executions")
+  @annotation.nowarn("msg=possible missing interpolator")
   lazy val githubWorkflowBuildConcurrency = settingKey[Option[JobConcurrency]](
     "Use concurrency to ensure that only a single build job within the same concurrency group will run at a time. (default: ci-${{ matrix.os }}-${{ matrix.java }}-${{ matrix.scala }}-${{ github.ref }})")
```
armanbilge, line 56: Instead of the `nowarn` this is the trick we usually use, but I don't have a strong opinion about it.

```suggestion
  lazy val githubWorkflowBuildConcurrency = settingKey[Option[JobConcurrency]](
    s"Use concurrency to ensure that only a single build job within the same concurrency group will run at a time. (default: ci-$${{ matrix.os }}-$${{ matrix.java }}-$${{ matrix.scala }}-$${{ github.ref }})")
```

.github/workflows/ci.yml
```diff
@@ -27,6 +27,9 @@ jobs:
         java: [temurin@8, temurin@17, graal_latest@17]
         project: [sbt-typelevelJVM]
     runs-on: ${{ matrix.os }}
+    concurrency:
+      group: ci-${{ matrix.os }}-${{ matrix.java }}-${{ matrix.scala }}-${{ matrix.project }}-${{ github.ref }}
```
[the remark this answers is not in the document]
iRevive: `ci` prefix is only for readability purposes. We can strip it.

github-actions/src/main/scala/org/typelevel/sbt/gha/GenerativePlugin.scala
```diff
@@ -64,6 +67,17 @@ object GenerativePlugin extends AutoPlugin {
 
   import autoImport._
 
+  private object MatrixKeys {
+    val OS = "os"
+    val Scala = "scala"
+    val Java = "java"
```
armanbilge, line 73: Stupid micro-optimization 😜 adding `final` means the compiler will inline them at use-site.
```suggestion
    final val OS = "os"
    final val Scala = "scala"
    final val Java = "java"
```

.github/workflows/ci.yml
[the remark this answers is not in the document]
armanbilge: I think it's good for readability, in which case `ci-build-` seems the best actually.

iRevive:
I've addressed the issues

armanbilge:
Sorry, I was looking at the docs for this feature:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

I understand that we can set concurrency at both the job level (as done here) and the workflow level. Is there any reason we should prefer the job level?

It seems to me that doing this at the workflow level would mean we don't need to worry about the matrix keys. And it also means that the concurrency limit will apply to every job in the workflow, such as site generation, coverage, etc.

iRevive:
TIL there is a workflow-level concurrency option.

I found this [example](https://docs.github.com/en/actions/examples/using-concurrency-expressions-and-a-test-matrix):
```yaml
# This allows a subsequently queued workflow run to interrupt previous runs
concurrency:
  group: '${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}'
  cancel-in-progress: true
```

If I get it right, we would like to have a single running workflow per branch at max. Does it mean that the group should be `${{ github.workflow }}-${{ github.event.pull_request.head.label || github.head_ref || github.ref }}`?

Not sure about the `github.event.pull_request.head.label` though. My brain barely works in the evening 😬

iRevive:
I made a few tests and `${{ github.workflow }} @ ${{ github.head_ref || github.ref }}` should be enough.

armanbilge:
Thanks! Do you know why we need `github.head_ref || github.ref`, in what situation is just `github.ref` not correct?

If not no worries, let's just merge and release a milestone so this can get some real-world testing :)

iRevive:
According to the [documentation](https://docs.github.com/en/actions/learn-github-actions/contexts), even `github.ref` should be enough:
> The fully-formed ref of the branch or tag that triggered the workflow run. For workflows triggered by `push`, this is the branch or tag ref that was pushed. For workflows triggered by `pull_request`, this is the pull request merge branch. For workflows triggered by `release`, this is the release tag created. For other triggers, this is the branch or tag ref that triggered the workflow run. This is only set if a branch or tag is available for the event type. The ref given is fully-formed, meaning that for branches the format is `refs/heads/<branch_name>`, for pull requests it is `refs/pull/<pr_number>/merge`, and for tags it is `refs/tags/<tag_name>`. For example, `refs/heads/feature-branch-1`.

Using `github.head_ref || github.ref` can lead to a situation when workflows triggered by `push` and `pull_request` events will not work along, the last one will always be prioritized.

armanbilge:
So then should it just be `github.ref` ?

armanbilge (approved):
Fantastic, thanks for all of your work on this, much appreciated ❤️

iRevive:
Yes, let's use `github.ref`. If we can have one more milestone or RC tag, we can verify on several projects that the configuration works as intended.

armanbilge:
Yes, there will be at least a few more milestones before 0.5.0 final release ... still more to do before it's ready.

No need for a clean commit history here :)

PR #437, opened 2022-12-20
Add Inkuire for scaladoc

zetashift:
For scala version 3.2.1+.

I have very little knowledge of sbt, so open to any critique!

### DIFF
--- a/settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
+++ b/settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
@@ -199,7 +199,7 @@ object TypelevelSettingsPlugin extends AutoPlugin {
       scalaVersion.value match {
         case V(V(3, 2, Some(build), _)) if build >= 1 =>
           Seq("-Ygenerate-inkuire")
-        case V(V(3, minor, _, _)) if minor >= 2 =>
+        case V(V(3, minor, _, _)) if minor >= 3 =>
           Seq("-Ygenerate-inkuire")
         case _ =>
           Seq.empty
--- a/settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
+++ b/settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
@@ -194,6 +194,14 @@ object TypelevelSettingsPlugin extends AutoPlugin {
           }
         }
     },
+    Compile / doc / scalacOptions ++= {
+      scalaVersion.value match {
+        case V(V(3, minor, Some(build), _)) if minor >= 2 && build >= 1 =>
+          Seq("-Ygenerate-inkuire")
+        case _ =>
+          Seq.empty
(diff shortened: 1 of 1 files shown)


armanbilge (approved):
> I have very little knowledge of sbt

Oh, you too? 😛

LGTM :) I wonder if there's a reason they didn't enable it by default, hopefully it doesn't break everyone's builds 😂

zetashift:
Force pushed a new commit, that just changes the author because I forgot to configure that :p

settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
```diff
@@ -194,6 +194,14 @@ object TypelevelSettingsPlugin extends AutoPlugin {
           }
         }
     },
+    Compile / doc / scalacOptions ++= {
+      scalaVersion.value match {
+        case V(V(3, minor, Some(build), _)) if minor >= 2 && build >= 1 =>
```
TonioGela, line 199: wait a sec, it wouldn't be applied in case of `3.x.0` even with x > 3, right? For example 3.3.0 falls out of this condition.

TonioGela: I would add something like
```scala
case V(V(3, 2, Some(build), _)) if build >= 1 => Seq("-Ygenerate-inkuire")
case V(V(3, minor, ), _)) if minor >= 2 => Seq("-Ygenerate-inkuire")
```
or some alternative syntax to express both cases

zetashift: Ah yeah, true! Thank you for spotting that

armanbilge: Woops, thanks! :)

TonioGela: 🙇

settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
```diff
@@ -195,8 +195,11 @@ object TypelevelSettingsPlugin extends AutoPlugin {
         }
     },
     Compile / doc / scalacOptions ++= {
+      // Enable Inkuire for Scala 3.2.1+
       scalaVersion.value match {
-        case V(V(3, minor, Some(build), _)) if minor >= 2 && build >= 1 =>
+        case V(V(3, 2, Some(build), _)) if build >= 1 =>
+          Seq("-Ygenerate-inkuire")
+        case V(V(3, minor, _, _)) if minor >= 2 =>
```
armanbilge, line 202:
```suggestion
        case V(V(3, minor, _, _)) if minor >= 3 =>
```

armanbilge (approved):
LGTM, take 2 😜

PR #440, opened 2022-12-29
Bump actions to Node 16 versions

rossabaker:
I try to avoid major bumps of dependencies in minor versions, but 0.4 will eventually cease to function without this.

https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout current branch (full)
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           fetch-depth: 0
 
@@ -48,7 +48,7 @@ jobs:
 
       - name: Setup Java (temurin@8)
         if: matrix.java == 'temurin@8'
-        uses: actions/setup-java@v2
+        uses: actions/setup-java@v3
         with:
           distribution: jdkfile
           java-version: 8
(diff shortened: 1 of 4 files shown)


armanbilge:
Thanks, do we know when that happens? Same as you, so I've been trying to roll this into sbt-typelevel 0.5.

rossabaker:
"By Summer 2023".

I would strongly consider putting this in 0.4, especially if it's as easy as this and 0.5 is a breaking change.

I'm hoping this also fixes these, which appear to be coming from the old setup-java:

```
Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-***-deprecating-save-state-and-set-output-commands/
```

There are similar warnings about `save-state` all over my builds, but those all link to the same `save-output` warnings.

armanbilge:
> I would strongly consider putting this in 0.4, especially if it's as easy as this and 0.5 is a breaking change.

Fair enough, I'll defer to your judgement here :)

rossabaker:
I can speak from bitter and ongoing experience about how frustrating it is when you're trying to patch a security vulnerability across multiple repositories and your build has rotted in the meantime (🖕🖕🖕 Jekyll 🖕🖕🖕).  Everything on a current 0.4 is doomed, but if we can do this in a patch:

- Many people will patch their way out of this before their builds rot and can do the 0.5 work at their leisure.
- Anyone caught with their pants down in 2H 2023 has an easy fix for this while they focus on their more urgent problem.

Since the only thing that changed in v3.0.0's of the actions is the Node version, I _think_ it's safe.  They presumably would have bundled any other breaking changes with that.

valencik:
I'm +1 on getting this in 0.4, the GitHub Actions functionality is a big draw for sbt-typelevel, so we should keep that in tip top shape.

armanbilge (approved):

Issue #373, opened 2022-09-19
Error publishing module with managed sources


valencik:
While trying to publish a new project with managed sources I hit the following error:
```
[error] java.util.NoSuchElementException: None.get
[error] 	at scala.None$.get(Option.scala:529)
[error] 	at scala.None$.get(Option.scala:527)
[error] 	at org.typelevel.sbt.TypelevelSettingsPlugin$.$anonfun$perConfigSettings$5(TypelevelSettingsPlugin.scala:281)
[error] 	at scala.collection.immutable.List.map(List.scala:293)
[error] 	at org.typelevel.sbt.TypelevelSettingsPlugin$.$anonfun$perConfigSettings$4(TypelevelSettingsPlugin.scala:281)

...

(parse / Compile / packageSrc / mappings) java.util.NoSuchElementException: None.get
```
([Full GH Actions CI run here](https://github.com/valencik/sequoia/actions/runs/3078849021/jobs/4975288216))

That's the `.get` here: https://github.com/typelevel/sbt-typelevel/blob/v0.5.0-M5/settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala#L281

And `show parse / Compile / packageSrc / mappings` in sbt yields the following files (paths slightly modified to remove extra details)

```
sequoia/modules/parse/src/main/scala/io/pig/sequoia/ParseBuddy.scala,io/pig/sequoia/ParseBuddy.scala
sequoia/modules/parse/src/main/scala/io/pig/sequoia/AntlrUtils.scala,io/pig/sequoia/AntlrUtils.scala
sequoia/modules/parse/src/main/scala/io/pig/sequoia/AstBuilder.scala,io/pig/sequoia/AstBuilder.scala
sequoia/modules/parse/target/scala-2.13/src_managed/main/antlr4/io/pig/sequoia/SqlBaseVisitor.java,antlr4/io/pig/sequoia/SqlBaseVisitor.java
sequoia/modules/parse/target/scala-2.13/src_managed/main/antlr4/io/pig/sequoia/SqlBaseParser.java,antlr4/io/pig/sequoia/SqlBaseParser.java
sequoia/modules/parse/target/scala-2.13/src_managed/main/antlr4/io/pig/sequoia/SqlBaseLexer.java,antlr4/io/pig/sequoia/SqlBaseLexer.java
[...]

PR #448, opened 2023-01-08
Add helpful error for misplaced managed sources

armanbilge:
Closes https://github.com/typelevel/sbt-typelevel/issues/373.

Just ran into this again, so now I've got a pretty good idea of what the user mistake is in this situation.

This error message would have helped in the original issue since we would have realized it's trying to mix Scala 2 and Scala 3 sources.


valencik (approved):
Nice! I think this would have helped past me.

Issue #450, opened 2023-01-17
Show published version in prominent location


bpholt:
When publishing snapshot versions in CI, it can be a little tricky to find the exact version being published in the build output. It's there, and in a consistent place, but the version doesn't stand out in a way that's easy to copy-and-paste. It would be nice to output it in a more prominent location.

One way to do this might be to use [GitHub Action Job Summaries](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/). I've also seen this done by printing the version in a semi-structured way in its own job step.

armanbilge:
This is a good one! Basically we want to add a step here that writes the value of `ThisBuild / version` to the file pointed at by the env variable `GITHUB_STEP_SUMMARY`.

https://github.com/typelevel/sbt-typelevel/blob/5d41b757e1e8099fba133060d7059f6161ef15d6/sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala#L62

TonioGela:
I'll take care of the implementation in #451 if you like 😇

PR #451, opened 2023-01-18
Populating Job summary with tests results and publish report

TonioGela:
This should address #450, or at least begin to address it.
The content that will appear in the file can be discussed (and made fancier than the mere version number)
It's the first time I'm contributing to an sbt plugin, so I'm not sure if eventual `IO` (the sbt one's ofc :D) errors should be handled safely and more precisely logged for example.


sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
```diff
@@ -57,9 +60,18 @@ object TypelevelSonatypeCiReleasePlugin extends AutoPlugin {
...
+      ) { summaryFile =>
+        val buildVersion: String = (ThisBuild / version).value
+        IO.write(new File(summaryFile), buildVersion)
+      }
+    },
     githubWorkflowTargetTags += "v*",
     githubWorkflowPublish := Seq(
-      WorkflowStep.Sbt(List("tlRelease"), name = Some("Publish"))
+      WorkflowStep.Sbt(List("tlRelease"), name = Some("Publish")),
+      WorkflowStep.Sbt(List("tlCreateJobSummary"), name = Some("Create Job Summary"))
```
armanbilge, line 74:
```suggestion
      WorkflowStep.Sbt(List("tlRelease", "tlCreateJobSummary"), name = Some("Publish"))
```

TonioGela: This config will result in `run: sbt '++ ${{ matrix.scala }}' tlRelease tlCreateJobSummary` that is not a valid syntax 🤔

sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
```diff
@@ -57,9 +61,17 @@ object TypelevelSonatypeCiReleasePlugin extends AutoPlugin {
 
       tags ++ branches
     },
+    tlCreateJobSummary := {
```
armanbilge, line 64: Unfortunately this name is a bit too vague—in sbt, all keys exist in global scope, so the name must be globally unique across all sbt plugins. That's why all the keys are prefixed with `tl` (typelevel), and then I also try to include the name of the specific sub-plugin. e.g. `tlCiReleaseStepSummary`, since this is the CI (sonatype) release plugin.

TonioGela: Makes sense

TonioGela:
😎 
picture

sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
```diff
@@ -44,6 +44,12 @@ object TypelevelSonatypeCiReleasePlugin extends AutoPlugin {
   override def globalSettings =
     Seq(tlCiReleaseTags := true, tlCiReleaseBranches := Seq())
 
+  private def renderSummaryTable(results: Map[String, String]): String =
+    results
+      .toList
+      .map { case (k, v) => s"| ${k} | ${v} |" }
+      .mkString(s"| Build Results | |\n| -: | :- |\n", "\n", "")
```
armanbilge: Since it's not specific to the build step?

```suggestion
      .mkString(s"| Workflow Results | |\n| -: | :- |\n", "\n", "")
```

TonioGela: It's hardcoded in the release step though, maybe `Release Workflow Results` is better? I'm wondering whether it will make sense to add this feature in a more agnostic way in to the `ghactions` plugin.

armanbilge: > I'm wondering whether it will make sense to add this feature in a more agnostic way in to the `ghactions` plugin. wdyt?

Yes me too. Hmm 😂

build.sbt
```diff
@@ -12,6 +12,10 @@ ThisBuild / developers := List(
 )
 ThisBuild / startYear := Some(2022)
 
+ThisBuild / tlCiReleaseStepSummaryTableInfo += (
+  "Scala Versions" -> crossScalaVersions.value.toString
+)
```
TonioGela, line 17: @armanbilge does it makes sense to keep it for this project? It's an sbt plugin after all...

ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
```diff
@@ -126,6 +126,39 @@ object TypelevelCiPlugin extends AutoPlugin {
...
+    Test / Keys.executeTests := {
+      val results: Tests.Output = (Test / Keys.executeTests).value
+      GitHubActionsPlugin.appendtoStepSummary(renderTestResults(Keys.name.value, results))
+      results
+    }
+  )
+
+  def renderTestResults(projectName: String, results: Tests.Output): String = {
+    val tableHeader: String =
+      s"### $projectName Tests Result\n" +
```
TonioGela, line 139: 🚲 🏠 : `Tests Result` or `Tests Results` or `Suite Results` ?

sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
```diff
@@ -57,9 +61,27 @@ object TypelevelSonatypeCiReleasePlugin extends AutoPlugin {
...
+      .mkString(s"| Build Results | |\n| -: | :- |\n", "\n", "\n")
+
+  override def projectSettings: Seq[Setting[_]] = Seq(
+    publish := {
+      GitHubActionsPlugin.appendtoStepSummary(
+        renderSummaryTable(tlCiReleaseStepSummaryTableInfo.value)
+      )
+      publish.value
+    }
+  )
```
TonioGela, line 86: I had no chance to really test it for obvious reasons

TonioGela:
Also, I still haven't added any documentation since features might be reworked according to PR feedback but I will do once the PR will be considered ready ofc

ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
armanbilge: They all seem fine. One other idea is to give the command that can be used to run that locally e.g.
`++scalaVersion projectName/test`

ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
```diff
@@ -126,6 +126,39 @@ object TypelevelCiPlugin extends AutoPlugin {
...
 
+  override def projectSettings: Seq[Setting[_]] = Seq(
+    Test / Keys.executeTests := {
+      val results: Tests.Output = (Test / Keys.executeTests).value
+      GitHubActionsPlugin.appendtoStepSummary(renderTestResults(Keys.name.value, results))
+      results
+    }
+  )
+
+  def renderTestResults(projectName: String, results: Tests.Output): String = {
```
armanbilge, line 137:
```suggestion
  private def renderTestResults(projectName: String, results: Tests.Output): String = {
```

build.sbt
armanbilge: Eh, we have a few things like this. dog-fooding various features within sbt-typelevel itself is a great way to test them. Esp since we don't have any real tests as you well know 😜

sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
```diff
@@ -57,9 +61,27 @@ object TypelevelSonatypeCiReleasePlugin extends AutoPlugin {
 
       tags ++ branches
     },
+    tlCiReleaseStepSummaryTableInfo := {
+      Map("**Release version**" -> (ThisBuild / version).value)
+    },
```
armanbilge, line 66: A couple other ideas for things we can put here:
1. The resolver it was published to (e.g. sonatype s01 snapshots), and how to configure it in sbt.
2. If populated, the value of `ThisBuild / apiURL`, since that will link to the scaladocs. Could be handy.

armanbilge (commented):
Overall looks great! Just a few thoughts :)

ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
TonioGela: Good Idea

sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
TonioGela: 2 is an easy task, but I'm not sure how to get the correct resolver in 1: the `resolvers` key is a `Seq[Resolver]` (and I bet is used just to resolve the deps I need to build my project, not as a target to publish to), and possibly the `publishTo` is just a project setting, not a build one right?

TonioGela: I'm not sure on how to proceed since:
1) `resolvers` is a `Seq[Resolver]` and it might not contain the resolver the project is using since it's used to define "user additional resolvers"

```scala
val resolvers = settingKey[Seq[Resolver]]("The user-defined additional resolvers for automatically managed dependencies.").withRank(BMinusTask)
```
There's `fullResolvers`

```scala
val fullResolvers = taskKey[Seq[Resolver]]("Combines the project resolver, default resolvers, and user-defined resolvers.").withRank(CTask)
```
but it's a task

2) `publishTo` might be the thing, but it's a task like `fullResolvers`, not a setting
3) `Resolver` is an abstract class that merely exposes a `name: String` attribute, so extracting the exact information to put in the instructions is not as trivial as desired.

TonioGela: Something like this might work
```scala
  override def projectSettings: Seq[Setting[_]] = Seq(
    publish := {
      val table: Map[String, String] = tlCiReleaseStepSummaryTableInfo.value
      val maybeResolver: Option[Resolver] = (ThisBuild / publishTo).value

      val textToRender: String =
        if (tlCiAddResolverInfoToSummary.value) {
          maybeResolver.fold(renderSummaryTable(table)) { r =>
            val text: String = renderSummaryTable(table + ("Resolver" -> r.toString))
            s"${text}\nTo configure your build to use this published version set\n`resolvers += ${r.toString}`"
          }
        } else renderSummaryTable(table)

      GitHubActionsPlugin.appendtoStepSummary(textToRender)

      publish.value
    }
  )
```

but requires an additional `tlCiAddResolverInfoToSummary` setting, plus it doesn't really solve the problem 3. We could write an exhaustive pattern match for resolvers, but do we really want to? :D

sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
```diff
@@ -57,9 +62,50 @@ object TypelevelSonatypeCiReleasePlugin extends AutoPlugin {
...
+      .mkString(s"| Build Results | |\n| -: | :- |\n", "\n", "\n")
+
+  override def projectSettings: Seq[Setting[_]] = Seq(
+    publish := {
+      val table: Map[String, String] = tlCiReleaseStepSummaryTableInfo.value
+      val maybeMavenResolverUrl: Option[(String, String)] =
+        (ThisBuild / publishTo).value.collect {
+          case x: MavenRepo => (x.name, x.root)
+          case x: MavenRepository => (x.name, x.root)
+        }
```
TonioGela, line 89: @armanbilge I tried this impl to get the resolver URL just in the cases the resolver is a maven repo with a URL, but it should be definitively tested

TonioGela: For the moment these settings are producing a bunch of 👇 

### sbt-typelevel-github-actions Publication Summary
| Build Result | Value |
| -: | :- |
| Release version | 0.4.18-24-5fe499e-20230205T100148Z-SNAPSHOT |
| Api URL | https://oss.sonatype.org/service/local/repositories/snapshots/archive/org/typelevel/sbt-typelevel-docs_2.12/0.4.18-24-5fe499e-20230205T100148Z-SNAPSHOT/sbt-typelevel-docs_2.12-0.4.18-24-5fe499e-20230205T100148Z-SNAPSHOT-javadoc.jar/!/index.html |
| Scala Versions | 2.12.17 |

sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
```diff
@@ -57,9 +62,52 @@ object TypelevelSonatypeCiReleasePlugin extends AutoPlugin {
...
+              val instructions: String =
+                s"To configure your build to use this published version set\n" +
+                  s"""`resolvers += Resolver.url("${n}", url("${u}"))`\n\n"""
+
+              renderSummaryTable(newTable) + instructions
+          }
+
+      GitHubActionsPlugin.appendtoStepSummary(header + textToRender)
+
+      publish.value
```
armanbilge, line 110: Hm, should we run this first, before populating the table with the results? You know, in case it fails 😂

sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
```diff
@@ -28,6 +29,10 @@ object TypelevelSonatypeCiReleasePlugin extends AutoPlugin {
       "Controls whether or not v-prefixed tags should be released from CI (default true)")
     lazy val tlCiReleaseBranches = settingKey[Seq[String]](
       "The branches in your repository to release from in CI on every push. Depending on your versioning scheme, they will be either snapshots or (hash) releases. Leave this empty if you only want CI releases for tags. (default: [])")
+    lazy val tlCiReleaseStepSummaryTableInfo = settingKey[Map[String, String]](
+      "Key-value set of information that will be rendered in a table in the job summary (default: [Release version -> ThisBuild/version])")
+    lazy val tlCiAddResolverInfoToSummary = settingKey[Boolean](
+      "If true adds the resolver url and instructions to the job summary (default: true)")
```
armanbilge, line 35: On second thought—do we _really_ need to expose these as settings? Basically it constrains us since we'd have to support these settings going forward. E.g. if we decided a different, non-tabular way to display the release data.

github-actions/src/main/scala/org/typelevel/sbt/gha/GitHubActionsPlugin.scala
```diff
@@ -120,6 +120,11 @@ object GitHubActionsPlugin extends AutoPlugin {
     }
   }
 
+  def appendtoStepSummary(summaryText: String) =
```
armanbilge, line 123: Let's make this private for now, we can use it internally in sbt-typelevel, and make it public if there's a usecase / request. This gives us a chance to tweak the API if we want to.
```suggestion
  private[typelevel] def appendtoStepSummary(summaryText: String) =
```

Besides, if someone _really_ wants this its easy to inline in their own plugin.

armanbilge (commented):
Nice job with the resolver stuff!

One last round of thoughts, this is where I start sweating nervously about compatibility guarantees 😅 basically lets think hard about what we want to expose yet—the less we expose, the more opportunity we have to iterate on this without breaking anyone.

sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
TonioGela: <you_saw_nothing_madagascar_meme.png> here

sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
TonioGela: Makes sense. It pairs well with the combatibility question you raised in the global comment. I'll hide those settings as mere private vals.

sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
TonioGela: > > as with `publishLocal` the publishTo doesn't get populated
> 
> I mean, that's fine, since running `publishLocal` in CI doesn't put it anywhere you can use it anyway :)

Not sure I follow. What I meant is that "on my laptop" the publishing fails before the command completes, and the only thing I had the chance to test is `publishLocal` overridden in the same way this `publish` is. To REALLY test with `publish` we should... well... publish :D


> I am also pretty sure it won't work on tagged releases which go to maven central. But that's also fine because everybody knows where to find those.

Hmm, where's the configuration for tagged releases? I might take a look to try to make it compatible.

armanbilge: > Hmm, where's the configuration for tagged releases? I might take a look to try to make it compatible.

Somewhere in https://github.com/xerial/sbt-sonatype

The reason I suspect it won't work is because sbt-sonatype does a "bundle release" when deploying to maven. So instead of setting `publishTo` to a sonatype resolver, IIUC it sets `publishTo` to your local file system, and then atomically uploads that as a bundle to Sonatype.

Anyway, I wouldn't worry about it. We can land this, and futz with that in a follow-up :)

TonioGela: > Anyway, I wouldn't worry about it. We can land this, and futz with that in a follow-up :)

Sounds good, we can handle these improvements atomically. Also, I saw that there are a bunch of enhancement proposals for this plugin that might deserve some attention now that the shade curtain between me and sbt plugins is getting thinner :D

I don't think I have nothing else to develop in this PR, I'll wait for the definitive review.

project/build.sbt
```diff
@@ -28,3 +28,6 @@ Compile / unmanagedResourceDirectories ++= modules.map { module =>
 libraryDependencies ++= Seq(
   "io.get-coursier" %% "coursier" % "2.1.0-RC5"
 )
+
+ThisBuild / libraryDependencySchemes ++= Seq(
+  "org.scala-lang.modules" %% "scala-xml" % VersionScheme.Always)
```
armanbilge: Huh, what changed that causes us to need this now?

armanbilge (approved):
Aight it's go time 😃 

Update: https://github.com/typelevel/sbt-typelevel/actions/runs/4141308116/attempts/1#summary-11245005824 😎

Ok the `***` thing is annoying, because we switched the org to `s01.oss.sonatype.org` via a secret, so GH is helpfully `***`ing that 🤔

TonioGela:
🎉 🎉 🎉 I'm so glad we merged it :)

PR #462, opened 2023-01-31
Add Laika API link configuration

reardonj:
This PR sets up the [Laika configuration for API links](https://planet42.github.io/Laika/0.18/03-preparing-content/02-navigation.html#linking-to-api-documentation), based on a map of package prefixes to scaladoc URLs provided by the plugin user. If none are provided, it will default to `tlSiteApiUrl` if available.

This allows links to the API docs to be easily added to Typelevel microsites, eg.: `@:api(cats.data.Chain)`

This implementation is based on the review comments for https://github.com/typelevel/sbt-typelevel/pull/225 and will supercede that PR.

### DIFF
--- a/site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
+++ b/site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
@@ -23,6 +23,8 @@ import laika.helium.config.Favicon
 import laika.helium.config.HeliumIcon
 import laika.helium.config.IconLink
 import laika.helium.config.ImageLink
+import laika.rewrite.link.ApiLinks
+import laika.rewrite.link.LinkConfig
 import laika.sbt.LaikaPlugin
 import laika.theme.ThemeProvider
 import mdoc.MdocPlugin
@@ -45,6 +47,8 @@ object TypelevelSitePlugin extends AutoPlugin {
     lazy val tlSiteHeliumConfig = settingKey[Helium]("The Typelevel Helium configuration")
     lazy val tlSiteHeliumExtensions =
       settingKey[ThemeProvider]("The Typelevel Helium extensions")
+    lazy val tlSiteApiLinks =
+      settingKey[Map[String, URL]]("URLs to API doc roots, by package prefix")
     lazy val tlSiteApiUrl = settingKey[Option[URL]]("URL to the API docs")
     lazy val tlSiteApiModule =
       settingKey[Option[ModuleID]]("The module that publishes API docs")
@@ -85,6 +89,7 @@ object TypelevelSitePlugin extends AutoPlugin {
   override def buildSettings = Seq(
     tlSitePublishBranch := Some("main"),
     tlSitePublishTags := tlSitePublishBranch.value.isEmpty,
+    tlSiteApiLinks := tlSiteApiUrl.value.map("*" -> _).toMap,
     tlSiteApiUrl := None,
     tlSiteApiPackage := None,
     tlSiteRelatedProjects := Seq(TypelevelProject.Cats),
@@ -106,6 +111,12 @@ object TypelevelSitePlugin extends AutoPlugin {
       .value: @nowarn("cat=other-pure-statement"),
     tlSitePreview := previewTask.value,
     Laika / sourceDirectories := Seq(mdocOut.value),
+    laikaConfig := laikaConfig
+      .value
+      .withConfigValue(LinkConfig(apiLinks = tlSiteApiLinks.value.toList.map {
+        case (packagePrefix, baseUri) =>
+          ApiLinks(packagePrefix = packagePrefix, baseUri = baseUri.toString)
+      })),
     laikaTheme := tlSiteHeliumConfig.value.build.extend(tlSiteHeliumExtensions.value),
     mdocVariables := {
       mdocVariables.value ++
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -64,6 +64,7 @@ Instead of using the super-plugins, for finer-grained control you can always add
 -  `TypelevelSitePlugin`: Sets up an [mdoc](https://scalameta.org/mdoc/)/[Laika](https://planet42.github.io/Laika/)-generated website, automatically published to GitHub pages in CI.
 - `tlSitePublishBranch` (setting): The branch to publish the site from on every push. Set this to `None` if you only want to update the site on tag releases. (default: `main`)
 - `tlSitePublishTags` (setting): Defines whether the site should be published on tag releases. Note on setting this to `true` requires the `tlSitePublishBranch` setting to be set to `None`. (default: `false`)
+- `tlSiteApiLinks` (setting): Map of package prefixes to API documentation sites for use with Laika API links. (default: * -> `tlSiteApiUrl`, if available)
 - `tlSiteApiUrl` (setting): URL to the API docs. (default: `None`)
 - `tlSiteHeliumConfig` (setting): the Laika Helium config. (default: how the sbt-typelevel site looks)
 


site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
```diff
@@ -85,6 +89,7 @@ object TypelevelSitePlugin extends AutoPlugin {
   override def buildSettings = Seq(
     tlSitePublishBranch := Some("main"),
     tlSitePublishTags := tlSitePublishBranch.value.isEmpty,
+    tlSiteApiLinks := tlSiteApiUrl.value.map("*" -> _).toMap,
```
armanbilge, line 92: Unfortunately the `tlSiteApiUrl` is not the "base" URL for the API. See (1) in https://github.com/typelevel/sbt-typelevel/pull/225#issuecomment-1077674413.

So maybe we need  to add a `tlSiteApiBaseUrl` or something ...

armanbilge: I also wonder if it would be helpful to populate this with some useful defaults. Like for Cats and Cats Effect and stuff, since I assume downstream documentation will want to reference it fairly frequently.

armanbilge:
For testing, can we try dog-fooding this in the sbt-typelevel documentation?

site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
```diff
@@ -106,6 +111,12 @@ object TypelevelSitePlugin extends AutoPlugin {
       .value: @nowarn("cat=other-pure-statement"),
     tlSitePreview := previewTask.value,
     Laika / sourceDirectories := Seq(mdocOut.value),
+    laikaConfig := laikaConfig
+      .value
+      .withConfigValue(LinkConfig(apiLinks = tlSiteApiLinks.value.toList.map {
+        case (packagePrefix, baseUri) =>
+          ApiLinks(packagePrefix = packagePrefix, baseUri = baseUri.toString)
+      })),
```
armanbilge, line 119: Does the order here matter? e.g. what happens if the `cats` prefix comes before the `cats.effect` prefix?

site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
reardonj: That's too bad, `tlSiteApiUrl` worked fine for Cats. Is there a reasonable situation where stripping off everything after the last `/` wouldn't work? Or only defaulting to `tlSiteApiUrl` if it ends in `/` ?

If `tlSiteApiBaseUrl` would only be used for `tlSiteApiLinks`, we may as well just have users populate `tlSiteApiLinks` themselves instead of adding yet another setting.

Is there a way to get those common library versions inside the plugin? I wouldn't want to embed specific versions in the defaults.

site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
reardonj: Order doesn't matter, [Laika uses longest prefix match](https://github.com/planet42/Laika/blob/605d8a46361abdd696e9e55fd74016e77538e8e7/core/shared/src/main/scala/laika/directive/std/LinkDirectives.scala#L58).

site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
armanbilge: > Is there a reasonable situation where stripping off everything after the last `/` wouldn't work?

Yeah, so for example http4s-dom sets its `tlSiteApiUrl` to this, so it "zooms" to the right package.
```
https://www.javadoc.io/doc/org.http4s/http4s-dom_sjs1_2.13/0.2.6/org/http4s/dom/index.html
```
So I don't think it would work there.

---

`tlSiteApiBaseUrl` can be used to make `tlSiteApiUrl ` and in many cases it will be auto-populated by the unidocs package. `tlSiteApiUrl` wasn't the right primitive.

---

> Is there a way to get those common library versions inside the plugin? I wouldn't want to embed specific versions in the defaults.

Ugh, good point, I forgot about this trickyness. Blah. We could try scouring the classpath for this information but it's a bit hacky. I have to think about this.

armanbilge: > Is there a way to get those common library versions inside the plugin?

Actually, we should try and use the same mechanism that sbt uses for API mappings. These are read out of the dependencies. Essentially Laika is replicating this same feature, we just need to glue them together. This might even be upstreamable into sbt-Laika.

reardonj: Is that sbt of package prefixes though?

armanbilge: It must be, or rather it must be aware of the fully-qualified-class-names. Because I can write `[[cats.Monad]]` in my scaladoc and it will auto-link it to the right version of the Cats scaladoc, without requiring any configuration on my part besides `autoAPIMappings := true`.

armanbilge: https://github.com/sbt/sbt/blob/3f1ff987905620718825b55d28cdcb65f094e7ce/main/src/main/scala/sbt/Defaults.scala#L2104

So it looks like sbt extracts a Map of Jar files to API urls. It's highly annoying, but it seems like basically want to index all those Jars to find out what packages are inside 😅

reardonj: That's a more ambitious project than I'm ready to bite off. We would still need `tlSiteApiBaseUrl` for the project docs, right? My understanding is that hypothetical machinery would only help for dependencies? And you want to revise `tlSiteApiUrl` to derive from the base url if it is available?

armanbilge: > We would still need `tlSiteApiBaseUrl` for the project docs, right

No, actually we wouldn't. This same hypothetical machinery should work for _everything_—that's why it is so appealing. For example in Cats you can see it includes both the dependencies and its own modules.

<details>

```
sbt:root> show docs/Compile/doc/apiMappings
Map(
  /modules/java.base -> https://docs.oracle.com/en/java/javase/11/docs/api/java.base/,
  /workspace/cats/kernel/.jvm/target/scala-2.13/classes -> https://oss.sonatype.org/service/local/repositories/snapshots/archive/org/typelevel/cats-docs_2.13/2.10-d15bc4b-SNAPSHOT/cats-docs_2.13-2.10-d15bc4b-SNAPSHOT-javadoc.jar/!/index.html,
  /home/gitpod/.cache/coursier/v1/https/repo1.maven.org/maven2/com/geirsson/metaconfig-core_2.13/0.9.10/metaconfig-core_2.13-0.9.10.jar -> https://github.com/olafurpg/metaconfig,
  /workspace/cats/core/.jvm/target/scala-2.13/classes -> https://oss.sonatype.org/service/local/repositories/snapshots/archive/org/typelevel/cats-docs_2.13/2.10-d15bc4b-SNAPSHOT/cats-docs_2.13-2.10-d15bc4b-SNAPSHOT-javadoc.jar/!/index.html,
  /home/gitpod/.cache/coursier/v1/https/repo1.maven.org/maven2/com/geirsson/metaconfig-typesafe-config_2.13/0.9.10/metaconfig-typesafe-config_2.13-0.9.10.jar -> https://github.com/olafurpg/metaconfig,
  /workspace/cats/free/.jvm/target/scala-2.13/classes -> https://oss.sonatype.org/service/local/repositories/snapshots/archive/org/typelevel/cats-docs_2.13/2.10-d15bc4b-SNAPSHOT/cats-docs_2.13-2.10-d15bc4b-SNAPSHOT-javadoc.jar/!/index.html,
  /home/gitpod/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scalacheck/scalacheck_2.13/1.16.0/scalacheck_2.13-1.16.0.jar -> https://www.javadoc.io/doc/org.scalacheck/scalacheck_2.13/1.16.0/,
  /workspace/cats/laws/.jvm/target/scala-2.13/classes -> https://oss.sonatype.org/service/local/repositories/snapshots/archive/org/typelevel/cats-docs_2.13/2.10-d15bc4b-SNAPSHOT/cats-docs_2.13-2.10-d15bc4b-SNAPSHOT-javadoc.jar/!/index.html,
  /home/gitpod/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.13.10/scala-library-2.13.10.jar -> https://www.scala-lang.org/api/2.13.10/,
  /workspace/cats/kernel-laws/jvm/target/scala-2.13/classes -> https://oss.sonatype.org/service/local/repositories/snapshots/archive/org/typelevel/cats-docs_2.13/2.10-d15bc4b-SNAPSHOT/cats-docs_2.13-2.10-d15bc4b-SNAPSHOT-javadoc.jar/!/index.html,
  /home/gitpod/.cache/coursier/v1/https/repo1.maven.org/maven2/org/typelevel/discipline-munit_2.13/2.0.0-M3/discipline-munit_2.13-2.0.0-M3.jar -> https://www.javadoc.io/doc/org.typelevel/discipline-munit_2.13/2.0.0-M3/,
  /home/gitpod/.cache/coursier/v1/https/repo1.maven.org/maven2/org/typelevel/discipline-core_2.13/1.5.1/discipline-core_2.13-1.5.1.jar -> https://www.javadoc.io/doc/org.typelevel/discipline-core_2.13/1.5.1/,
  /home/gitpod/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-compiler/2.13.10/scala-compiler-2.13.10.jar -> https://www.scala-lang.org/api/2.13.10/scala-compiler/,
  /home/gitpod/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-reflect/2.13.10/scala-reflect-2.13.10.jar -> https://www.scala-lang.org/api/2.13.10/scala-reflect/
)
```

</details>

> That's a more ambitious project than I'm ready to bite off.

No problem, sorry the scope exploded 😅 and anyway, now we are talking about a change in sbt-laika, since none of this is sbt-typelevel specific.

armanbilge: > This same hypothetical machinery should work for _everything_—that's why it is so appealing.

Oh wait, but I just realized it's pointing to the _snapshot_ scaladocs for Cats, not the latest stable version. Crap. It's not even wrong for doing so, since the docs are being built against the snapshot/dev version.

reardonj: But it should point to the right place for a actual publish builds?

armanbilge: Well, it depends how you are publishing the site.

If you are only publishing the site on releases, then it will point to the right place. But then you can't update the site between releases.

We are currently updating the site on every push to main, so technically it is built against an unreleased dev version.

There are well-known problems to both deploy strategies, so it is sort of a pick-your-poison ...

armanbilge: > So have we circled back to providing `tlSiteApiBaseUrl` then?

I think you also have to provide your package as well? Funny enough we might have an existing setting for that.

But I want to think about this more.

reardonj: …on the other hand, given the dream is to obsolete most of that config, and it's a one time bit of work on the project, maybe it's best not to add more machinery. here.

reardonj: > I think you also have to provide your package as well? Funny enough we might have an existing setting for that.

You can provide "*" as an explicit fallback (which is what the case class defaults to) and it handles that specially as a fallback if there is no match. So you don't need to provide your own prefix.

armanbilge: Right, but what happens if my project is `org.http4s.dom` and sbt-typelevel had pre-loaded `org.http4s` API docs for me. It just doesn't work in general.

reardonj: uhh; right

Issue #405, opened 2022-10-31
Strategically enable/disable bsp


armanbilge:
A really fantastic idea from @satorg 😃 

```scala
// project/BspControlPlugin.scala

object BspControlPlugin extends AutoPlugin {
  override def trigger = allRequirements
  override def requires = CrossPlugin

  override def projectSettings = Seq(
    Keys.bspEnabled :=
      CrossPlugin.autoImport.crossProjectPlatform.?.value.forall(_ == JVMPlatform)
  )
}
```

It's not right for every project, but this is a good default for the vast majority of projects. Also it's easy enough to opt-out with `.disablePlugins(BspControlPlugin)`.

I am not sure where in sbt-typelevel it should live. I guess it can go in the core plugin.

armanbilge:
I like the idea of a `tlBspPlatforms := Seq(JVMPlatform, NativePlatform)` setting.

satorg:
If I understand the overall idea correctly, we could have the `tlBspPlatforms` setting defined like this:
```scala
lazy val tlBspPlatforms = settingKey[Set[Platform]]("set of platforms for which BSP is enabled (default: JVMPlatform only)")
```
**Q**: `Set` or `Seq`? I like `Set` unless there's a concern about it.

then we could define a default value for it in the global scope:
```scala
override def globalSettings = Set( // or `buildSettings` ?
  tlBspPlatforms := Set(JVMPlatform)
)
```
Finally, we should change `bspEnabled` for each project based on `tlBspPlatforms`:
```scala
  override def projectSettings = Seq(
    Keys.bspEnabled :=
      // keep `bspEnabled := true` for non-crossproject projects
      CrossPlugin.autoImport.crossProjectPlatform.?.value.forall(tlBspPlatforms.value.contains(_))
  )
```

armanbilge:
@satorg yup, that sounds great to me. +1 to `Set`. Feel free to open a PR :)

I think the compatibility implications force us to put it in 0.5.x. We could put it in 0.4.x if we could default the `tlBspPlatforms` to all platforms, but we currently don't have a dependency on the Scala Native crossproject plugin. Maybe we should change that too at this point, lol.

satorg:
I've realized that I owed a PR for this issue. So here it is, finally: #465.

PR #465, opened 2023-02-06
BSP control plugin

satorg:
Addresses #405.
Note: targets the `main` branch currently!

Introduces  a new `tlBspPlatforms` setting of `Set[Platform]` type that controls for which cross-project platforms the `bspEnabled` setting should be set to `true`. By default it becomes enabled for `JVMPlatform` only.

The new setting can be useful to reduce BSP activity for re-compiling projects while editing their code files. Platforms like `JSPlatform` and `NativePlatform` usually  recompile slower and consume more resources (JVM heap, in particular) upon compilation. 

**How to change the default**

```scala
tlBspPlatforms := Set(JVMPlatform, JSPlatform)
```

**How to opt out**

```scala
.disablePlugins(BspControlPlugin)
```

### DIFF
--- a/core/src/main/scala/org/typelevel/sbt/BspControlPlugin.scala
+++ b/core/src/main/scala/org/typelevel/sbt/BspControlPlugin.scala
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2022 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.typelevel.sbt
+
+import sbt.Keys._
+import sbt._
+import sbtcrossproject.CrossPlugin
+import sbtcrossproject.CrossPlugin.autoImport._
+import sbtcrossproject.Platform
+
+object BspControlPlugin extends AutoPlugin {
+  override def trigger = allRequirements
+  override def requires: Plugins = CrossPlugin
+
+  object autoImport {
+    lazy val tlBspCrossProjectPlatforms: SettingKey[Set[Platform]] =
+      settingKey[Set[Platform]](
+        "set of platforms for which BSP should be enabled (default: not initialized)")
+  }
+
+  import autoImport._
+
+  override def projectSettings: Seq[Setting[?]] = Seq(
+    bspEnabled := {
+      val oldValue = bspEnabled.value
+
+      // We have to check if `crossProjectCrossType` and `crossProjectPlatform`
+      // are initialized. Otherwise, the `sbt-typelevel` won't load itself.
+      (
+        crossProjectCrossType.?.value,
+        crossProjectPlatform.?.value,
+        tlBspCrossProjectPlatforms.?.value
+      ) match {
+        // `tlBspCrossProjectPlatforms` is set by a user explicitly.
+        case (_, Some(projectPlatform), Some(bspPlatforms)) =>
+          bspPlatforms.contains(projectPlatform)
+        // If `CrossType` is `Pure` then enable BSP for the JVM platform only.
+        case (Some(CrossType.Pure), Some(projectPlatform), None) =>
+          projectPlatform == JVMPlatform
+        // Otherwise simply return the old value.
+        case _ => oldValue
+      }
+    }
+  )
+}
--- a/core/src/main/scala/org/typelevel/sbt/TypelevelBspPlugin.scala
+++ b/core/src/main/scala/org/typelevel/sbt/TypelevelBspPlugin.scala
@@ -22,7 +22,7 @@ import sbtcrossproject.CrossPlugin
 import sbtcrossproject.CrossPlugin.autoImport._
 import sbtcrossproject.Platform
 
-object BspControlPlugin extends AutoPlugin {
+object TypelevelBspPlugin extends AutoPlugin {
   override def trigger = allRequirements
   override def requires: Plugins = CrossPlugin
 


satorg:
Created as "Draft" coz I would like to do some additional checks first. But any feedback is welcome!

core/src/main/scala/org/typelevel/sbt/BspControlPlugin.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  import autoImport._
+
+  override def globalSettings: Seq[Setting[?]] = Seq(
+    tlBspPlatforms := Set(JVMPlatform)
+  )
+
+  override def projectSettings: Seq[Setting[?]] = Seq(
+    // Keep `bspEnabled := true` for non-sbtcrossproject projects.
+    bspEnabled := crossProjectPlatform.?.value.forall(tlBspPlatforms.value.contains)
```
armanbilge, line 43: This is good, but to confirm my understanding this plugin will only be enabled for `CrossProject`s anyway right? Because of the `requires: Plugins = CrossPlugin`.

core/src/main/scala/org/typelevel/sbt/BspControlPlugin.scala
```diff
@@ -0,0 +1,45 @@
...
+  object autoImport {
+    lazy val tlBspPlatforms: SettingKey[Set[Platform]] =
+      settingKey[Set[Platform]](
+        "set of platforms for which BSP is enabled (default: JVMPlatform only)")
+  }
+
+  import autoImport._
+
+  override def globalSettings: Seq[Setting[?]] = Seq(
+    tlBspPlatforms := Set(JVMPlatform)
```
armanbilge, line 38: I wish we didn't have to favor the `JVMPlatform` so blatantly here. It makes sense for many builds, but can also be less-than-ideal for some builds.

1. Projects defined as e.g. `crossProject(JSPlatform)`. Yeah this is dumb, but I see it all the time.
2. Cross-projects that have significant non-JVM sources. E.g. CE core, fs2-io.
3. The mythical `crossProject(JSPlatform, NativePlatform)`. I don't think I've ever seen this in the wild.

armanbilge (commented):
Thanks, I think we are fairly desperate for some relief in this area. Not your fault of course—but I just wish this wasn't so "clunky" for the non-vanilla usecases.

I wonder if it's worth going upstream to the `CrossPlugin` to try and expose some additional information e.g. all of a projects cross targets, whether it is a pure/full cross, etc.

Alternatively, I wonder if this is a problem that IDEs should solve, instead of us. Arguably it's a bit weird to hard-code BSPs into the build. Which BSPs should be enabled or disabled depends precisely on what the _user_ is doing at that moment, so it seems like they should configure this via UI in the IDE. To give an example: http4s-core is 99% a pure cross-build, but it has a very small amount of JS-specific sources. On the occasional bad day I have to poke at those, and it would be extremely annoying to have to futz the build to get BSP support. But it would be even worse to force JS BSP onto everyone because of like 3 source files I touch every 6 months when a new Node.js drops and everything breaks /rant.

satorg:
> Alternatively, I wonder if this is a problem that IDEs should solve, instead of us.

Yeah, I wish if it was on IDEs too. Unfortunately neither IntelliJ nor VSCode seem to provide such kind of filtering.

> I wonder if it's worth going upstream to the CrossPlugin to try and expose some additional information e.g. all of a projects cross targets, whether it is a pure/full cross, etc.

I think I can go ahead and create an issue in the **sbt-crossproject** repo and link this PR as a possible solution example. So we will see if there is any interest in such a feature. Wdyt, does it make sense to give it a try?

armanbilge:
I opened https://github.com/scalameta/metals-feature-requests/issues/332.

core/src/main/scala/org/typelevel/sbt/BspControlPlugin.scala
satorg: Yes, that's correct (sorry for long reply). So it should be pretty much safe simply to reference it via
```
crossProjectPlatform.value
```
I'm just not sure if it is possible to have `CrossPlugin` required but `crossProjectPlatform`.
Strictly saying, it is all up to the `CrossPlugin` implementation.
So I think maybe it makes sense to handle such a possibility, just in case.

core/src/main/scala/org/typelevel/sbt/BspControlPlugin.scala
satorg: changed to `tlBspExcludePlatforms` with empty set by default.

satorg:
Finally, I've made the setting less invasive and verified all the cases I was concerned about.
I think it is pretty much ready to go (if we decide it makes any practical sense, of course).

However, there's a potential catch with this approach (as a follow up of [this](https://discord.com/channels/632277896739946517/940325169057107968/1090073708648996928) discussion):

if we define the following combination of settings:
```sbt
ThisBuild / tlBspExcludePlatforms += NativePlatform // disable BSP in all projects for the "Native" platform
LocalProject("fooJS") / bspEnabled := false // also disable for the "JS" platform in the "foo" project specifically
```
then such a combination won't work.
It seems that the former setting gets a higher priority rather than the latter regardless of their order.

In other words, if `tlBspExcludePlatforms` is set to a non-default value, then further adjusting `bspEnabled` has no effect.

The issue can be solved though in this way:
```sbt
ThisBuild / tlBspExcludePlatforms += NativePlatform
LocalProject("eeeJS") / tlBspExcludePlatforms += JSPlatform
```
So the new setting shouldn't expose any blocker at the last.

armanbilge:
Thanks Sergey! Will take a closer look soon. Did you happen to see https://github.com/portable-scala/sbt-crossproject/pull/152, can we still make use of that here?

satorg:
@armanbilge sorry, I overlooked your question.
But I am not sure if that change in `sbt-crossproject` is somehow related, isn't it?
As I can see, it ultimately introduces two new settings: `crossProjectCrossType` and `crossProjectBaseDirectory`,
but I'm struggling to figure out how it could help to the functionality introduced in here.

Could you throw in a hint please?

armanbilge:
I thought at some point we had discussed using the `CrossType` to determine whether we only need BSP for one platform (because it's `Pure`) or for all platforms (because it's `Full`). But maybe that's no longer applicable?

armanbilge:
@satorg I think I'd like to move ahead with this, thanks for all your work! Did you have any thoughts on https://github.com/typelevel/sbt-typelevel/pull/465#issuecomment-1543387890?

satorg:
@armanbilge I added one more commit to the PR. It preserves `bspEnabled := false` if it was set that way somewhere else. Check it out please.

To you question:
> Did you have any thoughts on https://github.com/typelevel/sbt-typelevel/pull/465#issuecomment-1543387890?

Honestly, I'm not sure if it can be utilized here somehow.

Currently, the idea is that `BspControlPlugin` does _nothing_ by default. It only chimes in if `tlBspExcludePlatforms` is explicitly set to one or more platforms. So it does not make any own decisions.

I think, the best usage pattern of this plugin could be adding a separate file like `user.sbt` (or any other of one's preference) where to put a line like
```
ThisBuild / tlBspExcludePlatforms += NativePlatform
```
It should be a simple and quite harmless approach, I think.

Let me know what do you think please. Thank you!

armanbilge:
> Currently, the idea is that `BspControlPlugin` does _nothing_ by default.

Ah, I missed this. I thought it might be reasonable for the default to disable BSP for alternate platforms in pure crosses. The goal would be to lower the burden for casual contributors, who are unlikely to have a `user.sbt` configuration. Meanwhile regular contributors could invest in something like that.

armanbilge:
> So how can we figure out from the the `crossProjectCrossType` value that, let's say, `coreJVM` should have BSP enabled, whereas `coreJS` and `coreNative` should get it disabled?

We can have a setting for the "blessed" platform (defaults to JVM). then if it's a pure cross, it will only enable bsp if it's that blessed platform

... yeah, it's not great, but I think it would work in practice. idk, what do you think?

satorg:
@armanbilge, personally, I do not mind. But to be honest the "blessed platform" approach would work for me even for CrossProject.Full repos :) So I am not the best judge on that.

My only concern though is that if we take a look at a real big repo like Cats for an example, we will find out that roughly a half of its modules are CP.Pure whereas the second half is PC.Full. That means that the "blessed platform" approach will be only a half useful for Cats (or half useless, depending on your perception :) )

Also, maybe it'd make sense to add yet another settings: `tlBspIncludePlatforms`. I.e. if it happens that someone is not happy with the default logic, then they always can tackle that by adding an external `user.sbt` file with 
```
ThisBuild / tlBspIncludePlatforms += NativePlatform
```
Yeah, we can opt-out with `disablePlugins`, but it would require changes to the main `build.sbt`, whereas `tlBspIncludePlatforms` would allow to opt-out without such changes.

Let me know what you think please. Thanks!

armanbilge:
Ok, thanks for that exposition, I see what you are saying!

> I think, the best usage pattern of this plugin could be adding a separate file like `user.sbt` (or any other of one's preference) where to put a line like

I propose a slight revision to this. The best usage pattern of this plugin is that:

1. the `build.sbt` sets some reasonable default for project (possibly per-module), to optimize for best experience for typical/casual contribution

2. the power user adds a `user.sbt` with additional customization

Does that sound about right?

With that in mind can we just have one setting `tlBspPlatforms`? I guess we should default it to all 3. Big projects will have to opt-out, but I think thats better than non-JVM projects being confused.

satorg:
> With that in mind can we just have one setting tlBspPlatforms?

Yes, I think it would work even better (because looks simpler). The only caveat though, I think it should be set to empty by default, which would mean that it is not configured and the default logic applies per se. Once it is set to some non-empty value, it should supersede the default logic.

armanbilge:
> The only caveat though, I think it should be set to empty by default, which would mean that it is not configured and the default logic applies per se.

Hmm, it feels weird to special-case an empty value. We could either use an `Option` for the key, or actually not set a default value at all—in sbt you can add `.?` when reading a key so you get an `Option` depending on whether it is set or not.

satorg:
> or actually not set a default value at all—in sbt you can add .? when reading a key so you get an Option depending on whether it is set or not.

Yes agree, I think it would work too (maybe even better).

> it feels weird to special-case an empty value.

it does indeed, but on the other hand, an empty value for `tlBspPlatforms` does not make any sense per se, does it? I mean, if `tlBspPlatforms` would be set to an empty value, then what would be do about it, how would we handle this?

armanbilge:
> I mean, if `tlBspPlatforms` would be set to an empty value, then what would be do about it, how would we handle this?

That means they don't want any BSP 😛 you are right, there are better ways to indicate that.

Also, I just realize this setting only works for `crossProject`. So maybe we should make that clear e.g. `tlBspCrossPlatforms` ?

Ok, enough bike-shed 😁  I'm okay with any of the approaches, even just the empty value. I trust your judgement!

core/src/main/scala/org/typelevel/sbt/BspControlPlugin.scala
```diff
@@ -0,0 +1,59 @@
...
+
+import sbt.Keys._
+import sbt._
+import sbtcrossproject.CrossPlugin
+import sbtcrossproject.CrossPlugin.autoImport._
+import sbtcrossproject.Platform
+
+object BspControlPlugin extends AutoPlugin {
+  override def trigger = allRequirements
+  override def requires: Plugins = CrossPlugin
```
armanbilge, line 27: Looks like the `CrossPlugin` is enabled for all projects anyway.

armanbilge (approved):
Thank you so much for all the iterations on this. This is great!

PR #466, opened 2023-02-10
Making mergifyLabelPaths overridable

### LINKED ISSUE #220: Can't disable mergify auto-labeling
Setting `ThisBuild / mergifyLabelPaths := Map.empty` doesn't disable labeling.

TonioGela:
Closes #220.

### DIFF
--- a/mergify/src/main/scala/org/typelevel/sbt/mergify/MergifyPlugin.scala
+++ b/mergify/src/main/scala/org/typelevel/sbt/mergify/MergifyPlugin.scala
@@ -64,7 +64,8 @@ object MergifyPlugin extends AutoPlugin {
   override def buildSettings: Seq[Setting[_]] = Seq(
     mergifyStewardConfig := Some(MergifyStewardConfig()),
     mergifyRequiredJobs := Seq("build"),
-    mergifyLabelPaths := Map.empty,
+    mergifyLabelPaths := internallyCalculatedMergifyLabelPaths.value,
+    internallyCalculatedMergifyLabelPaths := Map.empty,
     mergifySuccessConditions := jobSuccessConditions.value,
     mergifyPrRules := {
       val baseDir = (LocalRootProject / baseDirectory).value.toPath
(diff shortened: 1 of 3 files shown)


TonioGela:
Hmm, running `mergifyGenerate` annihilates `.mergify.yml`, so the pr logic is probably wrong 👉 👈

mergify/src/main/scala/org/typelevel/sbt/mergify/MergifyPlugin.scala
```diff
@@ -114,8 +115,8 @@ object MergifyPlugin extends AutoPlugin {
       .dependsOn(ThisBuild / mergifyGenerate)
       .value,
     githubWorkflowCheck := githubWorkflowCheck.dependsOn(ThisBuild / mergifyCheck).value,
-    ThisBuild / mergifyLabelPaths := {
-      val labelPaths = (ThisBuild / mergifyLabelPaths).value
+    internallyCalculatedMergifyLabelPaths := {
```
armanbilge, line 118: You still need the `ThisBuild` :)
```suggestion
    ThisBuild / internallyCalculatedMergifyLabelPaths := {
```

TonioGela: 😎

bpholt:
> Hmm, running `mergifyGenerate` annihilates `.mergify.yml`, so the pr logic is probably wrong 👉 👈

I wonder if we ought to have some `scripted` tests for things like this, to show that the generated file is correct in different scenarios. (Not that you _have_ to add that to this PR if you don't want to, just that it might be helpful if you did.)

TonioGela:
I don't have that much experience with scripted tests if not in g8 templates, but AFAIK, the series/0.5 has some tests I might take inspiration from, I'll check and add them

armanbilge:
Yes, we _really_ ought to have tests in general. I've been lazy to do so, partly because testing generated output can become annoying. See some discussion in https://github.com/typelevel/sbt-typelevel/issues/26.

For this specific one, we can use a dog-food: override `mergifyLabelPaths` in `build.sbt`, just to prove we can. For example you should be able to remove this useless label.
- https://github.com/typelevel/sbt-typelevel/blob/abc217d7173ec51f44ed510e7d8cc5819543e7f2/.mergify.yml#L156-L163

armanbilge (approved):
Happy to close this one, finally! This was a good exercise in sbt setting evaluation and scoping 🤓

TonioGela:
> For this specific one, we can use a dog-food: override `mergifyLabelPaths` in `build.sbt`, just to prove we can. For example you should be able to remove this useless label.

I did it and it might act as a test as you said. For what concerns #26, probably the better thing is to implement them in a separate PR, innit?

PR #467, opened 2023-02-11
Use `-source:3.2-migration` on Scala 3.2 (when cross-compiling)

armanbilge:
By chance I stumbled on this flag today and h/t to @valencik and @samspills for concurrently discovering a scenario where it makes a difference in https://github.com/cozydev-pink/snakecase/pull/1.

Here is output of `scalac --help` for 3.2.2. Notice there is no `3.1-migration` option ...
```
              -source  source version
                       Default 3.2
                       Choices 3.0-migration, 3.0, 3.1, 3.2-migration, 3.2, future-migration, future
```


armanbilge:
I think we have to put this off until `series/0.5.x` 😢 

If you have
```scala
final case class Foo private (bar: Bar)
```

and you are using it like
```scala
Foo(bar)
```

This will stop compiling when this flag is added, because apparently under `3.2-migration` private constructor implies private `apply` method as well. Of course this is fixable, but it's really rough for an sbt-tl maintenance release.

PR #471, opened 2023-03-01
Removed default ligatures behaviour

j-mie6:
Set default font for code to Fira Mono, which doesn't have ligatures and added the `.ttf` files to resources to ensure they can be properly embedding in a generated `.pdf`


site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
```diff
@@ -146,6 +150,57 @@ object TypelevelSitePlugin extends AutoPlugin {
       tlSiteApiUrl.value.orElse(javadocioUrl).orElse(fallbackUrl)
     },
     tlSiteHeliumConfig := {
+      // default fontPath and fonts taken from Laika:
+      // instead of allowing for more fonts to be added, all the fonts must be respecified, to avoid "redundant embedding of unused fonts"
+      // as a result, these things have to just be defined again if we want to change them, and we do
+      val laikaFontPath = "laika/helium/fonts"
+      val tlFontPath = "org/typelevel/sbt/site/fonts"
```
armanbilge, line 157: Oh laika puts it under helium? I didn't think fonts are helium specific. Meh, we can leave it as is.

armanbilge (approved):
Thanks for chasing this down!! Ligatures begone!

site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
j-mie6: Well, we can't change it anyway right? This is where they are inside the laika jar

j-mie6: But yeah, I guess they put it there because the fonts are only used as "part of helium"

armanbilge: Right, I meant whether we should put it under helium too, to be consistent.

j-mie6: ohhhh

j-mie6: Easy change if you wanted it, but don't think it adds much really

armanbilge: Exactly. Meh :)

PR #472, opened 2023-03-03
Fixing section title sizing for code blocks

TonioGela:
It should fix this:
immagine

Testing it on sbt-typelevel itself resulted in this for an h1:
immagine


armanbilge:
Very happy to see this! WDYT about upstreaming this to Laika?

armanbilge:
Let's move ahead with this on `series/0.4`, and we can remove it in 0.5 when it upstreams. Thanks for chasing that!

armanbilge (approved):

PR #487, opened 2023-04-05
Clarify some tedious gotchas.

dragonfly-ai:
Reminder to add the .github/workflows/*.yml files to the git repository. A brief note about .scalafmt.conf for those who don't want to use the giter8 template.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -31,6 +31,12 @@ sbt new typelevel/typelevel.g8
 This will guide you through the basic setup to create a new project with **sbt-typelevel** and **sbt-typelevel-site**.
 Check out the [typelevel.g8](https://github.com/typelevel/typelevel.g8) project for more details.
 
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -31,6 +31,12 @@ sbt new typelevel/typelevel.g8
 This will guide you through the basic setup to create a new project with **sbt-typelevel** and **sbt-typelevel-site**.
 Check out the [typelevel.g8](https://github.com/typelevel/typelevel.g8) project for more details.
 
+Alternatively, if you don't want to use Giter8, you can add a file called: `.scalafmt.conf` to your project root directory with the contents:
```
armanbilge, line 34: I suggest we move this to the FAQ. Unfortunately adding a `.scalafmt.conf` by itself is not the only thing you need to do to completely avoid the giter 8, so I think this might be confusing :)

PR #489, opened 2023-04-09
Update sbt-header from 5.6.0 to 5.9.0

hnaderi:
Apparently, `sbt-header` `v5.6.0` is [no longer available](https://repo1.maven.org/maven2/de/heikoseeberger/sbt-header_2.12_1.0/).

### DIFF
--- a/core/build.sbt
+++ b/core/build.sbt
@@ -1,3 +1,3 @@
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.5.0")
-addSbtPlugin("de.heikoseeberger" % "sbt-header" % "5.6.0")
+addSbtPlugin("de.heikoseeberger" % "sbt-header" % "5.9.0")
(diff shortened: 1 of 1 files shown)


armanbilge:
That's because it is available from here:

https://scala.jfrog.io/ui/native/sbt-plugin-releases/de.heikoseeberger/sbt-header/scala_2.12/sbt_1.0/5.6.0/

which suffered a temporary blackout:
- https://github.com/sbt/sbt/issues/7202

If a published artifact ever disappears from Maven Central, that would be an extremely serious issue :)

---

Thanks for the PR. If I remember correctly I avoided updating from 5.6.0 because of bincompat breakage. Maybe to be conservative we should go to 5.6.3 on the 0.4.x branch. I need to look again at what changed.

We should definitely make sure we are on the latest sbt-header on the main branch that will become sbt-typelevel 0.5.

PR #501, opened 2023-05-03
use githubWorkflowBuildSbtStepPreamble in publish step

bpholt:
I think this may have been an oversight.

We don't have to change the default value of `sbtStepPreamble`, but with this change, the default value is never actually used anymore.

### DIFF
--- a/github-actions/src/main/scala/org/typelevel/sbt/gha/WorkflowJob.scala
+++ b/github-actions/src/main/scala/org/typelevel/sbt/gha/WorkflowJob.scala
@@ -20,7 +20,7 @@ final case class WorkflowJob(
     id: String,
     name: String,
     steps: List[WorkflowStep],
-    sbtStepPreamble: List[String] = List(s"++ $${{ matrix.scala }}"),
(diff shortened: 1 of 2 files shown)


github-actions/src/main/scala/org/typelevel/sbt/gha/WorkflowJob.scala
```diff
@@ -20,7 +20,7 @@ final case class WorkflowJob(
     id: String,
     name: String,
     steps: List[WorkflowStep],
-    sbtStepPreamble: List[String] = List(s"++ $${{ matrix.scala }}"),
+    sbtStepPreamble: List[String] = List.empty,
```
armanbilge, line 23: 👍 but probably better for 0.5.

bpholt:
Makes sense. I'm going to close this for now; we can retarget and resurrect it if/when it's ready.

armanbilge:
Thanks, ticketed here:
- https://github.com/typelevel/sbt-typelevel/issues/502

Issue #502, opened 2023-05-04
Dont do `++ scalaVersion` in publish job


armanbilge:
It's pointless.

armanbilge:
Taking this a step further: it should probably be possible to define `WorkflowJobs` without a scala axis, since not every job needs to cross on that at the matrix level.

PR #505, opened 2023-05-10
use githubWorkflowScalaVersions to set the Scala matrix for the validate-steward GHA job

bpholt:
This will clean up the resulting Action name, so that it doesn't change any time the Scala patch version changes, which is good for stable branch protection rules in GitHub.


.github/workflows/ci.yml
```diff
@@ -28,9 +28,12 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [2.12.17]
+        scala: [2.12]
         java: [temurin@8, temurin@17]
         project: [rootJVM]
+        exclude:
```
bpholt, line 34: I don't really understand why this exclusion was added—why is the default value based on `githubWorkflowScalaVersions.value.filterNot(_ == defaultScala)`?

https://github.com/typelevel/sbt-typelevel/blob/6077b3ed02762b853f0eed482af6c3f86c39c1a9/core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala#L79-L85

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.6.2
+sbt.version=1.8.2
```
bpholt, line 1: I think there was a reason this hasn't been updated, but I don't remember what it was. @armanbilge WDYT?

armanbilge:
Thanks! Actually I'd like to find a more general solution, which I mentioned briefly in https://github.com/typelevel/sbt-typelevel/issues/502#issuecomment-1535021360. It should be possible to omit the Scala axis completely from jobs, which would solve a lot of our problems.

bpholt:
> Thanks! Actually I'd like to find a more general solution, which I mentioned briefly in [#502 (comment)](https://github.com/typelevel/sbt-typelevel/issues/502#issuecomment-1535021360). It should be possible to omit the Scala axis completely from jobs, which would solve a lot of our problems.

That makes sense, but it'd have to wait for 0.5, wouldn't it? Could we merge this into 0.4?

armanbilge:
I think we could make that change compatibly. I believe currently an empty list for `scalas` in the job definition is not valid, but we could accept it now and use that to omit the scala version from the matrix.

bpholt:
That makes sense.

https://github.com/typelevel/sbt-typelevel/blob/6077b3ed02762b853f0eed482af6c3f86c39c1a9/github-actions/src/main/scala/org/typelevel/sbt/gha/GenerativePlugin.scala#L433-L437

I think we're basically saying if `job.scalas` is empty, then omit line 436. Should similar logic be applied to all of those matrix lines?

bpholt:
@armanbilge how do you feel about what I've done here?

.github/workflows/ci.yml
armanbilge: It's because you swapped `2.12.x` for `2.12`. Let's not make that change in 0.4.x, along with the sbt upgrade. It breaks too much stuff unfortunately, like this.

site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
```diff
@@ -310,7 +310,11 @@ object TypelevelSitePlugin extends AutoPlugin {
       WorkflowJob(
         "site",
         "Generate Site",
-        scalas = List((ThisBuild / scalaVersion).value),
+        scalas = List(
+          githubWorkflowScalaVersions
+            .value
+            .headOption
+            .getOrElse((ThisBuild / scalaVersion).value)),
```
armanbilge, line 317: This can be an empty `List(...)` now right?

github-actions/src/main/scala/org/typelevel/sbt/gha/GenerativePlugin.scala
```diff
@@ -443,6 +441,17 @@ ${indent(job.steps.map(compileStep(_, sbt, job.sbtStepPreamble, declareShell = d
...
   }
 
+  private def buildMatrix(level: Int, prefixWithEntries: (String, List[String])*): String =
+    prefixWithEntries
+      .map {
+        case (_, entries) if entries.isEmpty => none[String]
+        case (prefix, entries) =>
+          s"$prefix:${compileList(entries, level + 1)}".some
+      }
+      .collect { case Some(s) => s }
```
armanbilge, line 451:
```suggestion
      .collect {
        case (prefix, entries) if entries.nonEmpty =>
          s"$prefix:${compileList(entries, level + 1)}"
      }
```

site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
armanbilge: Oh, we'll have to clear the `sbtStepPreamble` as well.

Probably should be safe to make this similar kind of change for the publish job 🤔

bpholt: I wasn't sure if it mattered when generating sites (I haven't tried using the site plugin yet). I'll make it empty and we'll find out :)

.github/workflows/ci.yml
bpholt: Bummer… I think that makes it impossible to dogfood one of the actual things this is intended to change. I don't want to go deeper into a rabbit hole though so I'll revert that part of it.

bpholt:
fixes #502

site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
bpholt: [Looks like it worked](https://github.com/typelevel/sbt-typelevel/actions/runs/4952791623/jobs/8859547230?pr=505)?

armanbilge (approved):
Thanks for bearing with me, I'm so happy that this is finally fixed!! Solves a lot of problems.

armanbilge:
I'll get this out into a release later today or first thing tomorrow.

armanbilge:
Oh, just realized there is one more place where which may be affected by this.

https://github.com/typelevel/sbt-typelevel/blob/90eaaa4691745960fbe5970e390997f42e0c9624/github-actions/src/main/scala/org/typelevel/sbt/gha/GenerativePlugin.scala#L875-L885

bpholt:
> Oh, just realized there is one more place where which may be affected by this. … `def expandMatrix`

Maybe I'm missing what you're getting at, but I think this is already doing the right thing. `expandMatrix` is called from two places: [`GenerativePlugin.scala:626`](https://github.com/typelevel/sbt-typelevel/blob/8950243c4606b82f67ea07e0ebf8caa4e3f8b1f2/github-actions/src/main/scala/org/typelevel/sbt/gha/GenerativePlugin.scala#L625), where it's passed `githubWorkflowScalaVersions`, and [`MergifyPlugin.scala:140`](https://github.com/typelevel/sbt-typelevel/blob/8950243c4606b82f67ea07e0ebf8caa4e3f8b1f2/mergify/src/main/scala/org/typelevel/sbt/mergify/MergifyPlugin.scala#L140), where it's passed `job.scalas` (where `job` is a `WorkflowJob` mapped from `githubWorkflowGeneratedCI`).

I pulled local snapshots of the plugin as it exists now into a couple of different projects and the generated `Download target directories` jobs look ok to me, but again maybe I'm missing something.

armanbilge:
> I pulled local snapshots of the plugin as it exists now into a couple of different projects and the generated `Download target directories` jobs look ok to me, but again maybe I'm missing something.

Yes, in practice issues are unlikely because the build job always specifies the OS/Java/Scala axes. My concern is if one of those was empty, or this `expandMatrix` ends up used somewhere in the future where those can be empty.

bpholt:
That makes sense. Actually setting `ThisBuild / mergifyRequiredJobs += "validate-steward"` immediately revealed the problem you were worried about :)

Should have an update soon.

armanbilge (approved):
Thanks for the fixup!!

PR #510, opened 2023-05-13
Migrate Laika to `0.19`

danicheg:
As a matter of fact, I haven't tried to build any downstream project's microsite but building of `sbt-typelevel` microsite works well. It would be good if someone double-checks the results.

### DIFF
--- a/docs/site.md
+++ b/docs/site.md
@@ -53,4 +53,4 @@ lazy val unidocs = project
 
 ### How can I customize my website's appearance?
 
-We refer you to the comprehensive [Laika manual](https://planet42.github.io/Laika/index.html) and specifically the [`laikaTheme` setting](https://planet42.github.io/Laika/0.18/02-running-laika/01-sbt-plugin.html#laikatheme-setting).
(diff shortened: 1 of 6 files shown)


danicheg:
Also, if @jenshalm could glance at this few-liner, It'd be lovely (apologize for pinging you tho).

jenshalm:
The migration had already happened on the main branch (for the 0.5 milestones only) quite some time ago: #377.

I don't think there is a plan to migrate Laika for an 0.4.x patch release since it is a breaking change. It also requires more changes (apart from making it compile), mainly the removal of a bunch of workarounds which are no longer necessary.

danicheg:
Uh-oh, shame on me. Thanks!

armanbilge:
Sorry for the confusion everyone 😕 and also that the upgrade hasn't been released forever. My plan is to release sbt-typelevel 0.5 shortly after Scala 3.3 LTS. We'll enable a bunch of the new compiler flags which may break some builds, so this way we can condense the upgrade process.

PR #511, opened 2023-05-13
Pin Laika updates

danicheg:
Some follow-up to https://github.com/typelevel/sbt-typelevel/pull/510 😞


armanbilge (approved):

jenshalm:
Given that 0.18 won't receive any further updates and that the `main` branch should actually receive all 0.19 updates, wouldn't it be better to keep the old configuration and instead just change the target branch whenever scalasteward opens a PR for a Laika update?

armanbilge:
> wouldn't it be better to keep the old configuration and instead just change the target branch whenever scalasteward opens a PR for a Laika update?

Yeah, this is what I've been doing, but it seems to have caused some confusion :) nevermind, just a short transition period.

Issue #419, opened 2022-11-14
Add `scalac-compat` dependency by default


armanbilge:
Along with the kind-projector and bm4 plugins I think it's proving be essential.

See also:
- https://github.com/typelevel/sbt-typelevel/issues/375

PR #518, opened 2023-05-25
Add `scalac-compat` dependency by default

armanbilge:
Closes https://github.com/typelevel/sbt-typelevel/issues/419.


satorg (approved):
Thank you!

PR #523, opened 2023-06-04
Cleanup after scalafix update

mzuehlke:
The organize imports rule is now buildin.
Follow up to: #522 
See https://github.com/scalacenter/scalafix/releases/tag/v0.11.0


armanbilge (approved):
Oh, excellent, thank you!

Issue #474, opened 2023-03-04
Restore `-source:3.2-migration` 


armanbilge:
- https://github.com/typelevel/sbt-typelevel/pull/467
- https://github.com/typelevel/sbt-typelevel/pull/473

armanbilge:
Btw, I think the lesson learned here is we should try and get out the new flags ahead of Scala releases. Since upgrading a scala (minor) version might always involve work anyway, might as well as do any flag-related work at the same time. But _after_ everyone has upgraded it's extremely annoying to have to make more fixes because of a retro-actively applied flag.

armanbilge:
This seems pretty nuanced ...
- https://github.com/lampepfl/dotty/discussions/17591

armanbilge:
Conclusion in https://github.com/lampepfl/dotty/discussions/17591#discussioncomment-6003589 is that actually we should not use migration flags at all.

satorg:
Do you mean, this issue is better to be closed?

armanbilge:
We can close it after we remove the existing migration flag 😅 
https://github.com/typelevel/sbt-typelevel/blob/5efebdf5350f4e419970ae6906bea385d8381e3f/settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala#L173

Issue #257, opened 2022-04-11
`tlFatalWarnings` and `tlJdkRelease` don't respect `Config`, `Test`, etc. scopes


armanbilge:
Had to revert my initial attempt at this, see:
- https://github.com/typelevel/sbt-typelevel/pull/255

armanbilge:
I'm hoping that my proposal in https://github.com/typelevel/scalac-options/issues/2 will make this easier to implement.

armanbilge:
Depends on https://github.com/typelevel/sbt-typelevel/issues/256.

PR #515, opened 2023-05-25
Add re-introduced unused/discard warnings for scala >3.3.0

amumurst:
This PR prepares the plugin for the official scala 3.3.0 release (it is tagged, just not released quite yet https://github.com/lampepfl/dotty/releases/tag/3.3.0). It adds back all the warnings for unused and discarded values that is re-added in this release.

PR #493, opened 2023-04-19
Add -Xsource:3 to 2.12 (and future minor 2 releases) builds

etspaceman:

PR #531, opened 2023-06-18
Refresh scalac settings

armanbilge:
- Closes https://github.com/typelevel/sbt-typelevel/issues/474
- Merges https://github.com/typelevel/sbt-typelevel/pull/493
- Merges https://github.com/typelevel/sbt-typelevel/pull/515
- Heeds https://github.com/lampepfl/dotty/discussions/17591
- Closes https://github.com/typelevel/sbt-typelevel/issues/257

### DIFF
--- a/settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
+++ b/settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
@@ -269,7 +269,7 @@ object TypelevelSettingsPlugin extends AutoPlugin {
   private val perConfigSettings = Seq(
     scalacOptions := {
       val old = scalacOptions.value
-      val flag = "-Xfatal-warnings"
+      val flag = "-Werror"
       if (tlFatalWarnings.value)
         if (!old.contains(flag)) old :+ flag else old
       else
--- a/settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
+++ b/settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
@@ -233,20 +233,6 @@ object TypelevelSettingsPlugin extends AutoPlugin {
       "utf8",
       "-Xlint:all"
     ),
-
-    // TODO make these respect Compile/Test config
-    scalacOptions ++= {
-      if (tlFatalWarnings.value)
-        Seq("-Xfatal-warnings")
-      else
-        Seq.empty
-    },
-    javacOptions ++= {
-      if (tlFatalWarnings.value)
-        Seq("-Werror")
-      else
-        Seq.empty
-    },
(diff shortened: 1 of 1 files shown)


settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
```diff
@@ -281,6 +267,22 @@ object TypelevelSettingsPlugin extends AutoPlugin {
   ) ++ inConfig(Compile)(perConfigSettings) ++ inConfig(Test)(perConfigSettings)
 
   private val perConfigSettings = Seq(
+    scalacOptions := {
+      val old = scalacOptions.value
+      val flag = "-Xfatal-warnings"
+      if (tlFatalWarnings.value)
+        if (!old.contains(flag)) old :+ flag else old
+      else
+        old.filterNot(_ == flag)
```
satorg, line 276: A minor nit: `-Xfatal-warnings` was superseded with `-Werror` on 2.13 and 3.x. Although the former seems still working on both compilers. 

However, perhaps it makes sense to check for both options in the `if` condition, like
```
if (!old.contains(flag) && !old.contains("-Werror")) old :+ flag else old
else
  old.filterNot(opt => opt == flag || opt == "-Werror")
```

armanbilge: Oh, good to know 🤔 do you know which Scala versions that happened in?

I'm kind of okay with sbt-typelevel dropping support for old patches.

armanbilge: > However, perhaps it makes sense to check for both options in the `if` condition, like

The `if` is mainly intended to check for the flag it added itself, since `Test` seems to inherit from `Compile`, see https://github.com/typelevel/sbt-typelevel/pull/255#issue-1199911910.

satorg: @armanbilge 
> Oh, good to know 🤔 do you know which Scala versions that happened in?

looks like `-Werror` was added since 2.12.17:
```
$ scala-cli compile --server=false -S 2.12.16 -Werror -Xlint -- UnusedImport.scala 
scalac error: bad option: '-Werror'
  scalac -help  gives more information
Compilation failed

$ scala-cli compile --server=false -S 2.12.17 -Werror -Xlint -- UnusedImport.scala 
UnusedImport.scala:1: warning: Unused import
import scala.concurrent.Future
                        ^
error: No warnings can be incurred under -Xfatal-warnings.
one warning found
one error found
Compilation failed
```
However, seems it was not announced officially: https://github.com/scala/scala/releases/tag/v2.12.17.
Neither it is returned by any of scalac v2.12.17 help options (I couldn't find any at least).

PR #556, opened 2023-06-22
Add more JVMs to CI

durban:
This is just testing, because something doesn't seem to work with certain JVM versions.


build.sbt
```diff
@@ -14,8 +14,12 @@ ThisBuild / startYear := Some(2022)
 ThisBuild / githubWorkflowJavaVersions ++= Seq(
   JavaSpec.temurin("11"),
   JavaSpec.temurin("17"),
+  JavaSpec.graalvm("11"),
```
armanbilge, line 17: I'm not sure if GraalVM 11 is available like this. I think you need to specify it the "old" way.

https://github.com/typelevel/sbt-typelevel/blob/7a9074633fbb4d3fcf804f246c9f269885a5b35b/github-actions/src/main/scala/org/typelevel/sbt/gha/JavaSpec.scala#L43

armanbilge: Also is GraalVM for JDK 11 still relevant?

durban: `JavaSpec.graalvm("11")` seems to be the same as `JavaSpec(Distribution.GraalVM, "11")`.

armanbilge: Right, I think you need to do something like `JavaSpec(Distribution.GraalVM("21.3.0"), "11")`. This is the old versioning scheme, they changed it for the latest releases.

armanbilge: Ah, I see this is still broken 😅

durban: Yeah, apparently the generated yaml is invalid...

PR #562, opened 2023-06-24
docs for site plugin: adjust for recent changes to tlSiteIsTypelevelProject

jenshalm:
The docs did not yet reflect recent changes:

* The setting is no longer `Boolean`.
* It is now populated based on the `organization` setting.

### DIFF
--- a/docs/site.md
+++ b/docs/site.md
@@ -61,20 +61,20 @@ On top of the defaults of Laika standalone, sbt-typelevel-site adds:
 
 #### Additional Defaults for Typelevel Projects
 
-If the generated documentation is for a Typelevel project, you can optionally enable a set of additional defaults
-on top of the generic defaults listed in the previous section:
-
-```scala
-tlSiteIsTypelevelProject := true
-```
-
-With the flag above (which defaults to `false`) these additional settings apply:
+If the generated documentation is for a Typelevel project (meaning its `organization` setting is `org.typelevel`),
+a set of additional defaults will be enabled on top of the generic defaults listed in the previous section:
 
 * The home link in the top navigation bar carries the Typelevel logo and points to the Typelevel site.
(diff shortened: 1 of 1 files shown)


docs/site.md
```diff
@@ -61,21 +61,21 @@ On top of the defaults of Laika standalone, sbt-typelevel-site adds:
 
 #### Additional Defaults for Typelevel Projects
 
-If the generated documentation is for a Typelevel project, you can optionally enable a set of additional defaults
-on top of the generic defaults listed in the previous section:
-
-```scala
-tlSiteIsTypelevelProject := true
-```
```
armanbilge, line 69: Wouldn't it be helpful to keep an updated version of this? What I mean is, it seems far more likely an affiliate project would like to know how they can opt-in to the Typelevel theming, rather than an org project wanting to know how it can opt-out.

jenshalm: Yes, I just wasn't sure about the intention here. The `Organization` value is package-private and the `Affiliate` is not, yet in the implementation there is no difference between the two. I wasn't exactly sure what these design choices meant, so I did not write about it. 🙂

jenshalm: I added a section for `Affiliate` projects based on my somewhat limited understanding of the intentions. Feel free to rephrase or expand if necessary.

armanbilge: > yet in the implementation there is no difference between the two.

We decided in https://github.com/typelevel/governance/issues/34#issuecomment-1256693650 to start collecting this information even though we don't know what do with it yet, if anything.

> The `Organization` value is package-private and the `Affiliate` is not,

Yes, the idea is to discourage users from claiming they are an organization project. We'll see how it works 😀

armanbilge (approved):
Thank you!

PR #563, opened 2023-06-24
Scaladoc3 Wiki Syntax Flag

j-mie6:
Added the wiki syntax flag for non-pure Scala 3 projects, since scaladoc2 uses the wiki syntax with no markdown support. Pure Scala 3 builds may just use markdown support as intended.


armanbilge (approved):
thanks, this is great! had no idea we were missing this 😇

PR #570, opened 2023-06-26
Pin sbt to 1.9.x

armanbilge:


armanbilge:
The intent is to be conservative with sbt updates, so that updating sbt-typelevel doesn't force you to update your sbt version as well. If there is a compelling feature in a new sbt version we want to use here, we can always consider bumping and re-pinning.

mzuehlke:
Makes sense. thx

mzuehlke (approved):

danicheg:
It's been almost two years since this was pinned. Any plans to re-pin it to the 1.10 lineage?

armanbilge:
> Most crucially, the reason we might consider it relates to [SIP-51](https://docs.scala-lang.org/sips/drop-stdlib-forwards-bin-compat.html).

Hmm, what exactly is the relation? sbt-typelevel itself is using Scala 2.12. And just because it's pinned to sbt 1.9, doesn't prevent projects from using the latest sbt release.

danicheg:
@armanbilge ah, yes, sorry for the mislink. I actually meant 'initiatives like SIP-51', referring to the kind of impact they have (or would have) on the build. SIP-51 is definitely impactful, but it has nothing to do with Scala 2.12 indeed.

mzuehlke:
PR for updating to sbt 1.10.x #820

PR #573, opened 2023-06-26
New light theme for `sbt-typelevel-site`!

j-mie6:
Changed the default Helium theme over to one more clearly embodying the typelevel colours as found on the main website. Reworked the colours and syntax highlighting. Colours taken from typelevel are clearly labelled as such.

Here is an example:

image

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -2,7 +2,7 @@
 
 ## How do I cut a release?
 
-Create a release on GitHub with a v-prefixed, semantically-versioned tag (or, tag a commit locally and push to GitHub). This will start a CI release. Example tags: `v0.4.2`, `v1.2.3`, `v1.0.0-M1`, `v1.2.0-RC2`. 
+Create a release on GitHub with a v-prefixed, semantically-versioned tag (or, tag a commit locally and push to GitHub). This will start a CI release. Example tags: `v0.4.2`, `v1.2.3`, `v1.0.0-M1`, `v1.2.0-RC2`.
 
 It is also possible to run the release process entirely locally by invoking the `tlRelease` command, assuming that you have correctly configured your PGP keys and Sonatype credentials.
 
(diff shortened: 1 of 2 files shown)


rossabaker:
I like it, but I'm concerned about accessibility with the low contrast on the left.  It can be checked against [WCAG guidelines](https://webaim.org/resources/contrastchecker/).  Incidentally, the current version passes for large text and fails for small text, so this is a great chance to improve the status quo, too!

j-mie6:
I did check the syntax colouring for any concerns under firefox accessibility tools, and it was fine, didn't check side bar

j-mie6:
Yup, even `redTl`, (which actually looks a tad nicer imo) is a bit shy of the contrast requirements, I'll see if I can brighten the light pink a little more to fix it and will post an update screenshot :)

j-mie6:
image

@rossabaker hopefully that's a bit better

rossabaker (approved):
I think maybe I'd rather a code light theme to go with the rest of the light theme, but that's not really the point of this PR anyway.  This is a good idea and it looks like an improvement to me.

armanbilge:
FF4S has a nice light theme we might take inspiration from.
https://buntec.github.io/ff4s/examples/counter.html

j-mie6:
I can definitely make a code light theme, then can use this one for the (future) dark theme; light theme code highlighting is a bit trickier imo (weirdly!), So that can always come in a later PR (this doesn't need to be final, after all!

armanbilge (approved):
Personally I'm very excited about this. I can't thank you enough! 🥰 

@danicheg @satorg @ChristopherDavenport polite pings in case you have any opinions :)

danicheg (approved):
I like this 👍🏻 But it'd be good to check that this colour palette is also fine for people with visual impairments https://www.whocanuse.com/.

mzuehlke (approved):

j-mie6:
> I like this 👍🏻 But it'd be good to check that this colour palette is also fine for people with visual impairments https://www.whocanuse.com/.

I did run it through the Firefox impairment simulators at least and things seemed fine. Except for red-less colour blindness, that was fine but just... Bleak. That's to be expected, of course.

j-mie6:
That said, if any issues come up I'll be happy to find a fix for them :) (just @ me)

armanbilge:
@all-contributors add @j-mie6 for design

Issue #575, opened 2023-06-27
No-publish projects are recompiled during publishing ...


armanbilge:
In https://github.com/typelevel/sbt-typelevel/pull/534 we were clever and automatically omitted the no-publish projects from the artifact upload/download dance, since they would not be needed during the publish step.

However, when running MiMa during the publish job, it still attempts to compile these no-publish modules, before realizing that there is nothing to check binary-compatibility with. e.g. consider this log from Feral.

https://github.com/typelevel/feral/actions/runs/5386030675/jobs/9775733874#step:21:51

I think to fix this we may need to override the `mimaReportBinaryIssues` task so it no-ops if `publishArtifacts := false`. In fact, this is one of the tricky bits of sbt where we might need `taskDyn` type of thing, because the task dependency graph is changing dynamically based on the value of `publishArtifacts` i.e. `mimaReportBinaryIssues` may or may not depend on `compile`.

mzuehlke:
In the `TypelevelMimaPlugin` we are already overriding `mimaReportBinaryIssues`:
```Scala
    mimaReportBinaryIssues := {
      if (tlSkipIrrelevantScalas.value && (mimaReportBinaryIssues / skip).value)
        ()
      else mimaReportBinaryIssues.value
    },
```
and a bit later we are using `publishArtifact` while setting `mimaPreviousArtifacts`.

If we just combine this, it may already be enough.

mzuehlke:
Seems to have worked: https://github.com/typelevel/feral/actions/runs/5398011789/jobs/9803415627

PR #576, opened 2023-06-27
skip mima task, if no artifacts should be published

mzuehlke:
closes: #575


armanbilge (approved):
Ah, I was over-complicating things in the issue. This works perfectly, thank you!

PR #585, opened 2023-06-29
javadoc.io now supports sbt plugins

armanbilge:
Or more correctly, sbt plugins support Maven properly, so javadoc.io can pick up our docs and publish them.


mzuehlke:
This is only true, if a sbt > 1.9.0 is used. I know the RC1 release notes are mentioning this.
I tried to find a way to enforce this, ... with no success 😢

mzuehlke (approved):

armanbilge:
> I tried to find a way to enforce this, ... with no success

We could do it like this.
- https://github.com/ScalablyTyped/Converter/pull/540

PR #586, opened 2023-06-29
Landing page colour alterations

j-mie6:
Adjusted non-existent landing page colours to ensure they don't (hypothetically) look odd.

PR #590, opened 2023-06-30
New Theme Inversion

j-mie6:
Some people were unapply with the sheer in-your-face redness of the new theme introduced in #573, so this just inverts the primary and secondary colours, creating a bit better of a balance. With this, the lighter red is brought back, which is a bit more soft (but had bad contrast at the small sizes on the nav-bar), and the landing page changes from #586 have been incorporated.

However, the landing page in Laika is not particularly well set-up for light backgrounds, so while we wait for that to be properly decoupled (text colour-wise), there is an _**interim**_ darker landing page (note the button colour on that screenshot has since been altered):

image

image

Obviously, there is no default landing page for SBT typelevel currently, but this means the theme is currently a bit more future proofed. Eventually, the landing page will be lightened and all will be well with the world.

### DIFF
--- a/site/src/main/scala/org/typelevel/sbt/site/TypelevelSiteSettings.scala
+++ b/site/src/main/scala/org/typelevel/sbt/site/TypelevelSiteSettings.scala
@@ -26,6 +26,7 @@ import laika.theme.config.Color
 import org.typelevel.sbt.TypelevelGitHubPlugin.gitHubUserRepo
 import sbt.Def._
 import sbt.Keys.licenses
+import laika.ast.Path.Root
 
 object TypelevelSiteSettings {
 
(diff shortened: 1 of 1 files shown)


j-mie6:
With the merging of #594 the interim dark landing colours have been removed in favour of:

image

jenshalm:
A few fairly minor things I noticed for the normal content pages:

* The `primaryMedium` used for borders and lines is very light and barely visible. If you want to keep this tint for the landing page gradient, you could create one that is a tad darker for the normal pages.

* Literals in syntax highlighting are very dark, I find them hard to read. Even if it's temporary I think it would be good to make it a bit brighter.

* The background for code blocks is similar, but different to `primary`. It _might_ look more harmonic if you use `primary` as syntax-bg, too

j-mie6:
Addressed some of those (kept code background the same, as it is the colour for the TL logo's font and the main text colour:

image

(Also removed some unneeded colours, but kept `redTl` just in case)

jenshalm:
Looks good to me now.

PR #591, opened 2023-07-01
Deprecate `tlSkipIrrelevantScalas`

armanbilge:
We learned in https://github.com/typelevel/Laika/pull/451 that setting `crossScalaVersions := Nil` on the root project seems to be the trick to getting weird builds with "holes" to work, instead of this hack that was often broken.

### DIFF
--- a/ci/src/main/scala/org/typelevel/sbt/CrossRootProject.scala
+++ b/ci/src/main/scala/org/typelevel/sbt/CrossRootProject.scala
@@ -18,6 +18,7 @@ package org.typelevel.sbt
 
 import org.typelevel.sbt.gha.GenerativePlugin.autoImport._
 import sbt._
+import Keys._
 
 /**
  * Simultaneously creates a root project, a Scala JVM aggregate project, a Scala.js aggregate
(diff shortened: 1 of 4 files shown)


mzuehlke (approved):
Looks legit (👍

armanbilge:
@RustedBones

> This looks to work in the sbt shell

Does it? It didn't seem to be working for me.

Unfortunately I think your usecase is unsupported after this change 😕 this PR fixes the behavior when you run a command for all Scala versions e.g. `+test`. But if you set the Scala version first e.g. `++ 3` and then run `test` sbt does not automatically filter projects that do not support Scala 3.

There are two ways to fix it:

1. (workaround) cross-build the `scalacheck` project for Scala 3 as well so that the build is satisfied. You can move all your sources to `scala-2` folder and just have an empty artifact for Scala 3. If you set `publish / skip := tlIsScala3.value` then at least it won't publish.

2. use sbt-projectmatrix which can properly handle these situations.

Sorry about that. sbt is just really broken for these kinds of builds ...

RustedBones:
No worries. your work on this plugin saved me a huge amount of time already. Thanks a lot for looking into it so quickly

PR #600, opened 2023-07-08
Added binary versions to `module-ignore` inputs

### LINKED ISSUE #597: Dependency submission includes spurious dependencies
My dependency submission includes a bunch of stuff that isn't actually in the dependency list:

image

Build definition: https://github.com/Daenyth/cats-effect-guava/blob/main/build.sbt

j-mie6:
Added the binary scala suffix to the end of the module names for `Submit Dependencies`. The result of the change can be seen in the CI file. Fixes #597

### DIFF
--- a/no-publish/src/main/scala/org/typelevel/sbt/NoPublishPlugin.scala
+++ b/no-publish/src/main/scala/org/typelevel/sbt/NoPublishPlugin.scala
@@ -30,10 +30,10 @@ object NoPublishPlugin extends AutoPlugin {
     publishLocal := {},
     publishArtifact := false,
     publish / skip := true,
-    Global / noPublishModulesIgnore ++= crossScalaVersions.value.map { v =>
+    Global / noPublishModulesIgnore ++= crossScalaVersions.value.flatMap { v =>
       // the binary versions are needed for the modules-ignore in Submit Dependencies
       // it's best to pick them up here instead of guessing in the CI plugin
-      s"${thisProjectRef.value.project}_${CrossVersion.binaryScalaVersion(v)}"
+      CrossVersion(crossVersion.value, v, CrossVersion.binaryScalaVersion(v)).map(cross => cross(thisProjectRef.value.project))
     }
   )
 }
--- a/no-publish/src/main/scala/org/typelevel/sbt/NoPublishPlugin.scala
+++ b/no-publish/src/main/scala/org/typelevel/sbt/NoPublishPlugin.scala
@@ -19,7 +19,7 @@ package org.typelevel.sbt
 import sbt._
 
 import Keys._
-import NoPublishGlobalPlugin.autoImport.noPublishModulesIgnore
+import NoPublishGlobalPlugin.noPublishModulesIgnore
 
 object NoPublishPlugin extends AutoPlugin {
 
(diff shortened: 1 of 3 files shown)


mzuehlke (approved):
To me that looks like the best solution

no-publish/src/main/scala/org/typelevel/sbt/NoPublishPlugin.scala
```diff
@@ -30,7 +30,11 @@ object NoPublishPlugin extends AutoPlugin {
     publishLocal := {},
     publishArtifact := false,
     publish / skip := true,
-    Global / noPublishInternalAggregation += thisProjectRef.value
+    Global / noPublishModulesIgnore ++= crossScalaVersions.value.map { v =>
+      // the binary versions are needed for the modules-ignore in Submit Dependencies
+      // it's best to pick them up here instead of guessing in the CI plugin
+      s"${thisProjectRef.value.project}_${CrossVersion.binaryScalaVersion(v)}"
```
armanbilge, line 36: Unfortunately this is not sufficiently general. You should do something like this.
https://github.com/typelevel/sbt-typelevel/blob/fac0ec25fd0644c29238c0de9dfffdc8004a308b/sonatype/src/main/scala/org/typelevel/sbt/TypelevelSonatypePlugin.scala#L78-L82

armanbilge (changes requested):

no-publish/src/main/scala/org/typelevel/sbt/NoPublishPlugin.scala
armanbilge: btw, to dog-food this we can add some more no-publish modules to sbt-typelevel's own build.

Maybe one enabling the Scala.js plugin, one disable cross versions (i.e. a java module) and another enabling full cross versions (i.e. a compiler plugin)

j-mie6: can the `crossVersion` be assumed to be the same across all the cross-versions?

armanbilge: Yes, I suppose that has to be the assumption, otherwise everything breaks down 😅

no-publish/src/main/scala/org/typelevel/sbt/NoPublishPlugin.scala
```diff
@@ -40,17 +46,14 @@ object NoPublishGlobalPlugin extends AutoPlugin {
   override def trigger = allRequirements
 
   object autoImport {
-    lazy val noPublishProjectRefs = settingKey[Seq[ProjectRef]]("List of no-publish projects")
+    private[sbt] lazy val noPublishModulesIgnore =
```
armanbilge, line 49: Oh, sorry one last thing: since this is `private` there's no point to put it in the `autoImport`, since nobody can use it from `build.sbt` anyway. We can just keep it in the object.

j-mie6: fair, I wasn't sure if that was to just keep the imports cleaner in the other parts, I'll change that right away :)

armanbilge (approved):
thanks!

PR #602, opened 2023-07-09
Added `test-internal` to the ignored configs

j-mie6:
Turns out, transitive dependencies of `test` are counted as `test-internal`, this is now ignored.


mzuehlke (approved):

armanbilge (approved):
thanks for investigating this one!

PR #609, opened 2023-07-14
Add global permissions

etspaceman:
See https://github.com/sbt/sbt-github-actions/pull/105


.github/workflows/ci.yml
```diff
@@ -14,6 +14,9 @@ on:
     branches: ['**', '!update/**', '!pr/**']
     tags: [v*]
 
+permissions:
+  id-token: write
```
armanbilge, line 18: I'm afraid this will bork everything 😅 maybe we can be lazy and just set this to read-write...

> If you specify the access for any of these scopes, all of those that are not specified are set to `none`.

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#overview

etspaceman: updated

armanbilge (commented):
Just a couple nits but looks good to go. I like how this turned out!

Note for posterity: we had some discussion on Discord about whether to explicitly set permissions for the build, publish, site, etc. jobs. The tricky thing is it's hard to know exactly what permissions are required, particularly if the user has added some custom workflow steps. It's a bit annoying because as soon as we specify a few permissions for the things we _know_ we need, we have to specify _all_ the permissions with no way of falling back to the defaults that the user has set in their repository settings. So it's not very compositional.

I'm not entirely convinced how much security benefit it would bring either. So I think opt-in is okay for now, and we can revisit in the future.

github-actions/src/main/scala/org/typelevel/sbt/gha/GenerativePlugin.scala
```diff
@@ -210,6 +210,49 @@ object GenerativePlugin extends AutoPlugin {
...
+
+  def compilePermissions(permissions: Option[Permissions]): String = {
+    permissions match {
+      case Some(perms) =>
+        val rendered = perms match {
+          case Permissions.ReadAll => " read-all"
+          case Permissions.WriteAll => " write-all"
+          case Permissions.None => " {}"
+          case x: Permissions.Specify =>
+            val map = x.asMap.map {
```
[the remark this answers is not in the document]
etspaceman: Done

etspaceman:
@armanbilge did one more optimization by moving the string rendering to the Permissions classes. If you don't like that approach, no worries, we can revert the commit.

github-actions/src/main/scala/org/typelevel/sbt/gha/PermissionScope.scala
```diff
@@ -18,16 +18,25 @@ package org.typelevel.sbt.gha
 
 import scala.collection.immutable.SortedMap
 
-sealed abstract class Permissions extends Product with Serializable
+sealed abstract class Permissions extends Product with Serializable {
+  def rendered: String
+  def value: String = s"permissions:$rendered"
```
armanbilge, line 23: Do these need to be public?

etspaceman: They sure don't. Updated.

armanbilge:
> If you don't like that approach, no worries, we can revert the commit.

Oh yeah, maybe just leave it the way it was. It was fine and I think following the existing style of the project.

etspaceman:
Reverted

armanbilge (approved):
thanks!

PR #611, opened 2023-07-16
Revert #508

etspaceman:
This causes an issue in GHE runners that may not use the default java version of 11.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -490,17 +490,17 @@ jobs:
       - name: Checkout current branch (fast)
         uses: actions/checkout@v3
 
-      - name: Setup Java (temurin@17)
-        id: setup-java-temurin-17
-        if: matrix.java == 'temurin@17'
+      - name: Setup Java (temurin@11)
+        id: setup-java-temurin-11
+        if: matrix.java == 'temurin@11'
         uses: actions/setup-java@v3
         with:
           distribution: temurin
-          java-version: 17
+          java-version: 11
           cache: sbt
 
       - name: sbt update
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -484,11 +484,25 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
+        java: [temurin@11]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout current branch (fast)
         uses: actions/checkout@v3
 
+      - name: Setup Java (temurin@17)
```
armanbilge, line 493: Actually there was a mistake here before I realized I could delete it. The matrix specifies 11 but the generated steps set up 17 ...

ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
```diff
@@ -169,13 +169,13 @@ object TypelevelCiPlugin extends AutoPlugin {
             "validate-steward",
             "Validate Steward Config",
             WorkflowStep.Checkout ::
+              WorkflowStep.SetupJava(List(JavaSpec.temurin("17"))) :::
```
armanbilge, line 172: ... probably easiest to fix that by just setting this to 11.

etspaceman: Updated

.github/workflows/ci.yml
etspaceman: Updated

armanbilge (approved):
Thanks, annoying, but more robust 👍

PR #612, opened 2023-07-16
Remove sbt update from steward job

etspaceman:


.github/workflows/ci.yml
```diff
@@ -513,10 +513,6 @@ jobs:
           java-version: 11
           cache: sbt
 
-      - name: sbt update
-        if: matrix.java == 'temurin@11' && steps.setup-java-temurin-11.outputs.cache-hit == 'false'
-        run: sbt '++ ${{ matrix.scala }}' reload +update
```
armanbilge, line 518: So the issue is that it's still trying to cache like sbt. I think we should have a helper to generate a java-setup step with caching disabled. That would remove this parameter and also the `sbt update` step.

etspaceman: oh hm, let me look at that

etspaceman: Alright, let's see how you like this version :D

github-actions/src/main/scala/org/typelevel/sbt/gha/WorkflowStep.scala
```diff
@@ -41,16 +41,20 @@ object WorkflowStep {
     UseRef.Public("actions", "checkout", "v3"),
     name = Some("Checkout current branch (fast)"))
 
-  def SetupJava(versions: List[JavaSpec]): List[WorkflowStep] = {
+  def SetupJava(versions: List[JavaSpec], includeSbt: Boolean = true): List[WorkflowStep] = {
```
armanbilge, line 44: Or maybe `enableSbtCaching` or something

```suggestion
  def SetupJava(versions: List[JavaSpec], enableCaching: Boolean = true): List[WorkflowStep] = {
```

armanbilge (approved):

PR #624, opened 2023-08-12
Disable the `byname-implicit` lint on 2.13

armanbilge:
Due to an interaction of a scalac bug with shapeless derivation.

- https://github.com/scala/bug/issues/12072
- https://github.com/typelevel/sbt-tpolecat/issues/33

Also it is currently blocking the fs2 migration (see https://github.com/armanbilge/fs2/pull/4).

I haven't turned my brain on enough to appreciate the intrinsic value of this lint. Are we losing much by disabling it?

Edit: I guess the PR explains it in https://github.com/scala/scala/pull/8590.

Edit 2: so the primary use of implicit conversions is for syntax. And some Cats ops do take the target as by-name. But applying syntax to a multi-line block? Too weird 😅 I don't think we're missing much.


satorg (approved):
Yes, that is a very pesky bug. Iirc, it only exerts itself when some macros are involved. But seems impossible to suppress in code anyway. Thank you.

PR #639, opened 2023-09-11
Add `-Wconf:cat=scala3-migration:s` on 2.13

armanbilge:
This silences warnings cause by using `-Xsource:3` flag on Scala 2.13.

It turns out that the default behavior `-Xsource:3` is a bit tricky: it changes 2.13 semantics to match Scala 3, but then it also (fatally) warns you about those semantic changes. By adding this silencing, we can opt-in to the Scala 3 semantics without being warned about it.

This should silence two annoying warnings I'm aware of:
1. The warning on `package object`s that inherit from super-classes/traits. Our attempts to workaround this have gone very wrong e.g. https://github.com/typelevel/fs2/issues/3293.

2. A new one in Scala 2.13.12 where `case class`es with `private` constructors now raise a warning "constructor modifiers are assumed by synthetic `apply` method". Which is actually exactly the behavior we want.

Thanks to Som Snytt for exposition and providing the appropriate `-Wconf` flag on the Scala Discord.


satorg (approved):
LGTM, thank you!

PR #640, opened 2023-09-11
`cat=scala3-migration` only exists in 2.13.12+

armanbilge:
Fixes https://github.com/typelevel/sbt-typelevel/pull/639.

### DIFF
--- a/settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
+++ b/settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
@@ -114,21 +114,25 @@ object TypelevelSettingsPlugin extends AutoPlugin {
         "-Ywarn-dead-code", // superseded by "-Wdead-code"
         "-Ywarn-numeric-widen" // superseded by "-Wnumeric-widen"
       )
-      val warnings213 = Seq(
+      def warnings213(patch: Int) = Seq(
         "-Wdead-code",
         "-Wextra-implicit",
         "-Wnumeric-widen",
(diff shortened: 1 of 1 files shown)


settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala
```diff
@@ -149,8 +153,8 @@ object TypelevelSettingsPlugin extends AutoPlugin {
         case V(V(3, _, _, _)) =>
           warningsDotty
 
-        case V(V(2, minor, _, _)) if minor >= 13 =>
-          (warnings211 ++ warnings212 ++ warnings213 ++ warningsNsc)
+        case V(V(2, minor, patch, _)) if minor >= 13 =>
+          (warnings211 ++ warnings212 ++ warnings213(patch.getOrElse(0)) ++ warningsNsc)
```
j-mie6, line 157: under what circumstances is the `0` generated here?

armanbilge: should be never. the issue is we are reusing the `V` parser which is also used for parsing base versions that don't have patches.

j-mie6: I'd be _tempted_ to crash explicitly instead, since this should never happen

j-mie6: For consistency with #643, perhaps we should pass `Option[Int]` into `warnings213` and use `patch.forall(_ >= 12)`

PR #643, opened 2023-09-15
Added version bound check and added slash

j-mie6:
Addresses https://github.com/scala/bug/issues/12867, ensuring that a slash is added for `2.13.12` and above (assuming that this will also be the case for `2.13.13` and `2.14.0` etc).

### DIFF
--- a/github/src/main/scala/org/typelevel/sbt/TypelevelGitHubPlugin.scala
+++ b/github/src/main/scala/org/typelevel/sbt/TypelevelGitHubPlugin.scala
@@ -93,7 +93,7 @@ object TypelevelGitHubPlugin extends AutoPlugin {
             infoOpt.toSeq flatMap { info =>
               val path =
                 // see https://github.com/scala/bug/issues/12867#issuecomment-1718481858
-                if (minor > 13 || minor == 13 && patch.exists(_ >= 12))
+                if (minor > 13 || minor == 13 && patch.forall(_ >= 12))
                   s"${info.browseUrl}/blob/${vh}/€{FILE_PATH}.scala"
                 else s"${info.browseUrl}/blob/${vh}€{FILE_PATH}.scala"
               Seq("-doc-source-url", path)
--- a/github/src/main/scala/org/typelevel/sbt/TypelevelGitHubPlugin.scala
+++ b/github/src/main/scala/org/typelevel/sbt/TypelevelGitHubPlugin.scala
@@ -18,12 +18,12 @@ package org.typelevel.sbt
 
 import com.github.sbt.git.SbtGit.git
 import org.typelevel.sbt.kernel.GitHelper
+import org.typelevel.sbt.kernel.V
 import sbt._
 
 import scala.util.Try
 
 import Keys._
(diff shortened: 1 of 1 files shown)


github/src/main/scala/org/typelevel/sbt/TypelevelGitHubPlugin.scala
```diff
@@ -82,19 +82,24 @@ object TypelevelGitHubPlugin extends AutoPlugin {
...
+          case V(V(3, _, _, _)) =>
+            userRepo.toSeq flatMap {
+              case (user, repo) =>
+                Seq(s"-source-links:github://${user}/${repo}", "-revision", vh)
+            }
+          case (V(V(2, minor, patch, _))) =>
+            infoOpt.toSeq flatMap { info =>
+              val path =
+                // see https://github.com/scala/bug/issues/12867#issuecomment-1718481858
+                if (minor > 13 || minor == 13 && patch.exists(_ >= 12))
```
j-mie6, line 96: The behaviour here is to assume that someone running on `2.13` (whatever that means), would want the mitigation off

```suggestion
                if (minor > 13 || minor == 13 && patch.forall(_ >= 12))
```

This would instead assume that a `2.13`er would instead be running the latest version: is this what we want?

j-mie6:
I've checked this against `parsley`, as per usual: it correctly links in both `2.13.11` and `2.13.12`

github/src/main/scala/org/typelevel/sbt/TypelevelGitHubPlugin.scala
```diff
@@ -82,19 +82,24 @@ object TypelevelGitHubPlugin extends AutoPlugin {
...
+                  s"${info.browseUrl}/blob/${vh}/€{FILE_PATH}.scala"
+                else s"${info.browseUrl}/blob/${vh}€{FILE_PATH}.scala"
```
armanbilge, line 98: Dumb question: what if we just always put the slash? Does it work with 2.12, old 2.13s?

j-mie6: I wondered about this, then I didn't want to play with 🔥

j-mie6: Conclusion is, while it does double the slash in the link, this _appears_ to be benign and the link still gets to github ok. However, that feels dodgy to me

github/src/main/scala/org/typelevel/sbt/TypelevelGitHubPlugin.scala
armanbilge: Based on https://github.com/typelevel/sbt-typelevel/pull/643#discussion_r1327523924, I think we should fallback to adding the slash.

armanbilge (approved):

j-mie6:
Applied the desired change (first time using "apply suggestion", that was fun): we now fallback to slash if the patch is not specified.

PR #652, opened 2023-10-24
Introduce `tlCommandAliases` setting for defining command aliases

armanbilge:
If I may say so myself: this is a significant improvement for usability 🎉  it makes it much much easier to define command aliases that depend on settings, and furthermore to composably modify the command aliases downstream.


satorg (approved):
Yes, that seems neat, thanks!

PR #659, opened 2023-11-06
More robust configuration of `unmanagedSourceDirectories`

armanbilge:
Previously, we were forgetting to configure `scala-2.13+` and `scala-2.12-` for the non-`shared` directories of `CrossProject`s. h/t @j-mie6 for reporting.

So in this PR I'm trying a strategy that is a lot dumber but simpler to implement and more robust to unusual setups. It simply looks for all the source directories that end in `/scala` and adds `/scala-{2.12-,2.13+}` variants of them.

Notably, we no longer need to know about `CrossProject`s at all.

Please somebody tell me this is a bad idea 😂 it feels just a little bit too hacked. But I honestly cannot think of a better way to do this, esp. when there is also the explosion of `jvm-native` shared directories that also need to be considered ...

This PR also turns the `settings` plugin into a `CrossProject` so we can properly dogfood the changes.


j-mie6 (approved):
LGTM, I can test out a snapshot later if we want to be more safe that it works. Though it seems like we are changing the way the paths themselves are being constructed a bit, is that a problem cross-OS?

j-mie6:
> > is that a problem cross-OS?
> 
> CI passed on Windows and it exercises this code. So must be working?

Ok :)

j-mie6:
Test reveals that indeed, `scala-2.13+` is only added on the `jvm` project build, not for `native`, `js`, `shared`, or any combination of `js-jvm` etc.

armanbilge:
> Please somebody tell me this is a bad idea 😂 it feels just a little bit too hacked

So the hack part of this, where I think it's breaking down, is that our configuration must run _last_ after all the other source directories are configured. I don't think that's happening right now and I'm not really sure if we can guarantee that in general anyway.

This approach may be bust.

j-mie6 (changes requested):
Still needs to fix for the cross-target folders

PR #664, opened 2023-11-13
Improve tlSitePreview's console output

zetashift:
In my terminal (`wezterm+zellij`) the URL is clickable, not sure if I should test it on other terminals as well?

### DIFF
--- a/site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
+++ b/site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
@@ -256,7 +256,7 @@ object TypelevelSitePlugin extends AutoPlugin {
 
       val (_, cancel) = Tasks.buildPreviewServer.value.allocated.unsafeRunSync()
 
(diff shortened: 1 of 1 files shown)


site/src/main/scala/org/typelevel/sbt/TypelevelSitePlugin.scala
```diff
@@ -256,7 +256,7 @@ object TypelevelSitePlugin extends AutoPlugin {
 
       val (_, cancel) = Tasks.buildPreviewServer.value.allocated.unsafeRunSync()
 
-      logger.info(s"Preview server started on port ${laikaPreviewConfig.value.port}.")
+      logger.info(s"Preview server started on port http://localhost:${laikaPreviewConfig.value.port}.")
```
armanbilge, line 259: We're currently discarding the server in the `_`, we can actually get the address/port directly out of that instead of hard-coding.

armanbilge:
> URL is clickable, not sure if I should test it on other terminals as well?

URLs are clickable in my terminals as well 👍

valencik (approved):
nice

armanbilge (approved):
thank you!

PR #672, opened 2023-12-13
Allow packaging sources from all managed directories

RustedBones:
When using generators that put sources in custom directories, eg [sbt-avro](https://github.com/sbt/sbt-avro) where the generated sources are by default is a dedicated folder in `sourceManaged / avro / $config` instead of `sourceManaged / $config`.

Using `sbt-typelevel` fails the `packageSrc` with this setup.

As the custom managed source path is added to the `managedSourceDirectories`, check the path given from the mapping is part of one of the source instead of the default `sourceManaged`.


armanbilge (approved):
Thanks for the fix!

Issue #680, opened 2024-02-20
Install sbt with brew on macOS 13, 14


armanbilge:
For some reason GHA won't include it by default like they do in all the other images 😕 

- https://github.com/actions/runner-images/issues/9369#issuecomment-1953963333

PR #681, opened 2024-02-22
Install sbt via brew on macOS runners

armanbilge:
Fixes https://github.com/typelevel/sbt-typelevel/issues/680. To keep things simple we do the install for all macOS runners.


mzuehlke (approved):
Good solution.

While looking at the generate `ci.yml` 
I see that all jobs except  "Build and Test" contain a lot of steps that will never be executed because of their guards, and all the other jobs are only running on "ubuntu".

armanbilge:
Thanks!

> While looking at the generate `ci.yml`
> I see that all jobs except "Build and Test" contain a lot of steps that will never be executed because of their guards, and all the other jobs are only running on "ubuntu".

Yes this is currently the case because we use `githubWorkflowJobSetup` for all jobs. If we are clever we could probably filter irrelevant steps. In practice, it doesn't matter too much besides the bloat/noise.

PR #685, opened 2024-03-01
Run scalafix before scalafmt in `prePR`

satabin:
Since scalafix rewrites source code, the formatting might need some fixing afterwards.


armanbilge (approved):
great catch, thanks!

PR #698, opened 2024-04-18
Make `startYear` setting deterministic

armanbilge:
Instead of defaulting to the current year (this was a bad default, I'm sorry) it now defaults to the current year at the time that version of sbt-typelevel was built.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,7 +10,7 @@ ThisBuild / developers ++= List(
   tlGitHubDev("ChristopherDavenport", "Christopher Davenport"),
   tlGitHubDev("djspiewak", "Daniel Spiewak")
 )
-ThisBuild / startYear := Some(2022)
+ThisBuild / startYear ~= { _.map(_ => 2022) } // dog food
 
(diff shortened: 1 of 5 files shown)


mzuehlke:
Do you think even having a default for `startYear` is a good idea in the first place ?
I think it is mostly (exclusively ?) used for the header generation and wouldn't this result in the header being updated each year when we drop a new sbt-tl release ?

The official meaning  is
```Scala
val startYear = settingKey[Option[Int]]("Year in which the project started.").withRank(BMinusSetting)
```
https://github.com/sbt/sbt/blob/0c6d7b8f300f1e405e2a2db5d84f72d30a7ec901/main/src/main/scala/sbt/Keys.scala#L342

armanbilge:
> Do you think even having a default for `startYear` is a good idea in the first place ?

That's a good point. I think I gave it a default to make the "getting started" experience as fast and easy as possible. But I guess the proper place to do this is in templates where `startYear` set appropriately to a fixed year.

PR #700, opened 2024-04-20
Support `working-directory` in `WorkflowStep.Run`

armanbilge:
- https://github.com/sbt/sbt-github-actions/issues/179


armanbilge:
Btw, this is an example of the evolution that is effectively impossible to do compatibly without the changes in https://github.com/typelevel/sbt-typelevel/pull/694.

PR #706, opened 2024-05-06
Bump peaceiris/actions-gh-pages to 4.0.0

TonioGela:
Bumps the dep so that the warning about Node v16 being deprecated (when the site gets published) will disappear.
See https://github.com/peaceiris/actions-gh-pages/blob/v4.0.0/CHANGELOG.md#400-2024-04-08


TonioGela:
We just hit this in CI: https://github.com/actions/setup-java/issues/625

PR #711, opened 2024-05-16
Support checking Java formatting

armanbilge:
The sbt-java-formatter plugin exposes a `javafmtCheckAll` command.

https://github.com/sbt/sbt-java-formatter


mzuehlke:
Don't we need to add the plugin?

mzuehlke:
And what about the commands? Similar to scalafmt?

armanbilge:
> Don't we need to add the plugin?

The sbt-typelevel-ci plugin doesn't actually add the dependencies (same for scalafmt/scalafix/MiMa/etc.). Unfortunately we just need a centralized place to organize all the various commands so they are easy to toggle for users that do want to use those plugins.

> And what about the commands? Similar to scalafmt?

Sorry, I didn't understand this question.

mzuehlke (approved):

Issue #721, opened 2024-07-23
Publishing via the Central Portal


m3is0:
Since recently, all new Sonatype accounts must publish via the Central Portal:
[https://central.sonatype.org/register/central-portal/](https://central.sonatype.org/register/central-portal/)

I've successfully published my project with sbt-ci-release and sbt-sonatype v3.11.0 (which supports the new API), 
but with sbt-typelevel-ci-release v0.7.2 (which includes the same sbt-sonatype) it doesn't work.

I tried to set SONATYPE_CREDENTIAL_HOST, but with no success.

armanbilge:
Thanks for trying this!

> it doesn't work.

How do you know that it didn't work? Are there logs or some kind of error message?

m3is0:
As @kubukoz mentioned, I got the following error:
```
scala.MatchError: URLRepository(https://SettingKey(This / This / This / sonatypeCredentialHost), Patterns(ivyPatterns=Vector(), artifactPatterns=Vector(), isMavenCompatible=false, descriptorOptional=false, skipConsistencyCheck=false), false) (of class sbt.librarymanagement.URLRepository)
	at org.typelevel.sbt.TypelevelSonatypeCiReleasePlugin$.$anonfun$tlCiReleaseCommand$4(TypelevelSonatypeCiReleasePlugin.scala:75)
...
```

aartigao:
FYI, I've tried 👇🏽 

```Scala
addSbtPlugin("org.typelevel" % "sbt-typelevel" % "0.7.2-8-ce3d86e-SNAPSHOT")
```

and works! 🎉 Thanks @kubukoz!

(beware that, for the moment, publishing snapshots is not supported and your CI will fail).

kubukoz:
@aartigao did you identify where the problem with snapshots comes from?

aartigao:
If I read correctly it's just that it is not supported yet by SBT Sonatype.

m3is0:
It works well with ```SONATYPE_CREDENTIAL_HOST = "central.sonatype.com"``` as an actions secret!

With ```ThisBuild / sonatypeCredentialHost := xerial.sbt.Sonatype.sonatypeCentralHost``` it doesn't work.

It still uses the default host:
```
2024-07-27 20:14:52.597Z  info [SonatypeService] sonatypeRepository  : https://s01.oss.sonatype.org/service/local  - (SonatypeService.scala:26)
...
[...]

PR #722, opened 2024-07-23
Fallback on non-MavenRepository repos

kubukoz:
Currently, if you publish to the new Sonatype Central repository (#721), during publishing you'll encounter the following issue:

```
[error] scala.MatchError: URLRepository(https://settingkey/(This / This / This / sonatypeCredentialHost), Patterns(ivyPatterns=Vector(), artifactPatterns=Vector(), isMavenCompatible=false, descriptorOptional=false, skipConsistencyCheck=false), false) (of class sbt.librarymanagement.URLRepository)
```

It happens in the code that computes a step summary. Originally, I tried to write a rendering function for the entire `URLRepository`, but serializing `Patterns` is a bit tedious and the step summary isn't all that useful, so I figured we can start with an empty string for now.

### DIFF
--- a/sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
+++ b/sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
@@ -78,9 +78,11 @@ object TypelevelSonatypeCiReleasePlugin extends AutoPlugin {
                 |resolvers += "${repo.name}" at "${repo.root}"
                 |```
                 |""".stripMargin
+
+          case _ => ""
         }
 
-        GitHubActionsPlugin.appendtoStepSummary(
+        GitHubActionsPlugin.appendToStepSummary(
           s"""|## Published `$v`
               |${resolver}""".stripMargin
         )
--- a/sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
+++ b/sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
@@ -81,7 +81,7 @@ object TypelevelSonatypeCiReleasePlugin extends AutoPlugin {
 
           case repo: URLRepository =>
             s"""|```scala
-                |resolvers += Resolver.url("${repo.name}", ${repo.patterns}, ${repo.allowInsecureProtocol})
+                |resolvers += URLRepository("${repo.name}", ${repo.patterns}, ${repo.allowInsecureProtocol})
                 |```
                 |""".stripMargin
           // fallback to print at least _something_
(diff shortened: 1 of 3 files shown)


majk-p (approved):

zainab-ali (approved):

aartigao (approved):

aartigao:
> serializing `Patterns` is a bit tedious and the step summary isn't all that useful

While this is true, I think returning at least `URLRepository.name` will be better in this case. And I'll keep the empty string for global fallback.

kubukoz:
yo @armanbilge can we get this merged&released please? 🥺

armanbilge:
> I think returning at least `URLRepository.name` will be better in this case.

I agree. 

> the step summary isn't all that useful

I find it very useful :) I copy-paste from there to get the resolver config whenever I need to depend on a snapshot.

The goal is to print something there that will help the user.

armanbilge:
Btw, thanks so much for investigating this, very appreciated 🙏

kubukoz:
TBH I don't think that'll fix it.

```
URLRepository(https://SettingKey(This / This / This / sonatypeCredentialHost)
```

this would imply that there was a `toString` on a `SettingKey` used instead of a `.value` (or equivalent) :/

aartigao:
> TBH I don't think that'll fix it.

WDYM? At least the `MatchError` will be fixed, right? You mean that after fixing that it's still failing?

kubukoz:
> > TBH I don't think that'll fix it.
> 
> WDYM? At least the `MatchError` will be fixed, right? You mean that after fixing that it's still failing?

I haven't tried out the change yet, just thinking that if the `URLRepository` is constructed this way, and we use that to publish, then it's likely wrong.

But maybe it's not actually used for publishing and only used to print the result, in which case it'd be ok 😅

sonatype-ci-release/src/main/scala/org/typelevel/sbt/TypelevelSonatypeCiReleasePlugin.scala
```diff
@@ -78,9 +78,17 @@ object TypelevelSonatypeCiReleasePlugin extends AutoPlugin {
                 |resolvers += "${repo.name}" at "${repo.root}"
                 |```
                 |""".stripMargin
+
+          case repo: URLRepository =>
+            s"""|```scala
+                |resolvers += Resolver.url("${repo.name}", ${repo.patterns}, ${repo.allowInsecureProtocol})
```
armanbilge, line 84: Thanks, this is looking much better! But this API doesn't seem to exist 🤔 
https://github.com/sbt/librarymanagement/blob/8d71f66997c356790fc887b17e33c7be66a4e590/core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala#L350-L361

kubukoz: corrected to use `URLRepository.apply`

armanbilge (approved):
thanks for chasing this down!

PR #728, opened 2024-08-02
Add `with*` methods for `WorkflowJob`

zarthross:
- Added `with*` methods for `WorkflowJob` for symmetry with `WorkflowStep`

### DIFF
--- a/github-actions/mima.sbt
+++ b/github-actions/mima.sbt
@@ -1,0 +1,23 @@
+import com.typesafe.tools.mima.core._
+mimaBinaryIssueFilters ++= Seq(
+  "org.typelevel.sbt.gha.WorkflowJob.withId",
+  "org.typelevel.sbt.gha.WorkflowJob.withName",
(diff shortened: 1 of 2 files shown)


zarthross:
Mima is failing because I added methods.  Not sure why since this repo is early-semver...  Should I bump the minor version or add ignores?

armanbilge:
> Not sure why
- https://github.com/lightbend-labs/mima/issues/817

You can add ignores.

github-actions/mima.sbt
```diff
@@ -0,0 +1,23 @@
+import com.typesafe.tools.mima.core._
+mimaBinaryIssueFilters ++= Seq(
```
armanbilge, line 2: I think MiMa might support wildcards, but this is good too.

```scala
ProblemFilters.exclude[ReversedMissingMethodProblem](
  "org.typelevel.sbt.gha.WorkflowJob.with*"
)
```

armanbilge (approved):
Thank you!

Issue #204, opened 2022-03-09
Improve `prePR` and CI feedback, esp. for new contributors


armanbilge:
After an extensive discussion on [Discord](https://discord.com/channels/632277896739946517/632277897448652844/951144217504251955) about "What `prePR` means to you", I think we roughly came to this consensus:

1. `prePR` should only do polish tasks, which are essentially things that can be fixed by the task itself. Additionally, the `polish` check should be a separate job in CI, like suggested in https://github.com/typelevel/sbt-typelevel/issues/93.
2. We should try and relax the CI-requires-approval-for-new-contributors, since it really kills the feedback loop. I'm pretty sure there's a mode to only require it for new GH users, and I bet an admin can set this for the whole org.

Roughly, `prePR`/polish tasks should meet this criteria:
- it happens without any configuration or control
- it's infallible when it runs
- once you run it locally, the check in ci will never fail
- things that check something and fix it if needed on their own

As much as CI failures are annoying, at least maintainers can help. We don't want `prePR` to discourage contributors from creating a PR in the first place by failing with e.g. mima issues that they don't understand or know how to fix.

There's also a lot of interest in an idea to ask CI to run `prePR` for you like in https://github.com/typelevel/sbt-typelevel/issues/69, but it's possibly tricky to implement.

armanbilge:
See also https://github.com/typelevel/sbt-typelevel/pull/203#issuecomment-1063234479.

rossabaker:
Note that some scalafix are "polish" (e.g., organize imports), and some require intervention (e.g., no `fs2.Compiler[Sync]`).  And some of those that are "polish" are semantic rules, and therefore relatively slower.

PR #729, opened 2024-08-03
Improved prePR

zarthross:
I've found `prePR` to be a nice handy alias, but a little clumsy to extend.  My team has been doing something similar for a while, but did a command alias in every project.  I thought it would be nice if we had a handy easy way to extend the list of checks a developer should perform before submitting a pr so they have fewer painful moments realizing they forget a step and now their PR has failed.

### DIFF
--- a/core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
+++ b/core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
@@ -76,24 +76,30 @@ object TypelevelPlugin extends AutoPlugin {
         java <- githubWorkflowJavaVersions.value.tail // default java is head
       } yield MatrixExclude(Map("scala" -> scala, "java" -> java.render))
     },
-    GlobalScope / tlCommandAliases ++= {
+    GlobalScope / tlPrePRSteps ++= {
       val header = tlCiHeaderCheck.value
       val scalafmt = tlCiScalafmtCheck.value
       val javafmt = tlCiJavafmtCheck.value
       val scalafix = tlCiScalafixCheck.value
 
-      val prePR = List("project /", "githubWorkflowGenerate") ++
+      List("project /", "githubWorkflowGenerate") ++
         List("+headerCreateAll").filter(_ => header) ++
         List("+scalafixAll").filter(_ => scalafix) ++
         List("+scalafmtAll", "scalafmtSbt").filter(_ => scalafmt) ++
-        List("javafmtAll").filter(_ => javafmt)
+        List("javafmtAll").filter(_ => javafmt) ++
+        List("compile", "test")
+    },
+    GlobalScope / tlCommandAliases ++= {
+      val header = tlCiHeaderCheck.value
+      val scalafmt = tlCiScalafmtCheck.value
+      val javafmt = tlCiJavafmtCheck.value
 
       val botHook = List("githubWorkflowGenerate") ++
         List("+headerCreateAll").filter(_ => header) ++
+        List("javafmtAll").filter(_ => javafmt) ++
         List("+scalafmtAll", "scalafmtSbt").filter(_ => scalafmt)
 
       Map(
-        "prePR" -> prePR,
         "tlPrePrBotHook" -> botHook
       )
     }
--- a/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
+++ b/kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
@@ -45,6 +45,9 @@ object TypelevelKernelPlugin extends AutoPlugin {
       "Command aliases defined for this build"
     )
 
+    lazy val tlPrePRSteps =
+      settingKey[List[String]]("Steps to be performed before a user submits a PR")
+
     @deprecated(
       "Use `tlCommandAliases` for a more composable command definition experience",
       "0.6.1")
@@ -62,7 +65,8 @@ object TypelevelKernelPlugin extends AutoPlugin {
   override def globalSettings = Seq(
     Def.derive(tlIsScala3 := scalaVersion.value.startsWith("3.")),
     tlCommandAliases := Map(
-      "tlReleaseLocal" -> List("reload", "project /", "+publishLocal")
+      "tlReleaseLocal" -> List("reload", "project /", "+publishLocal"),
+      "prePR" -> ("reload" :: "project /" :: tlPrePRSteps.value)
     ),
     onLoad := {
       val aliases = tlCommandAliases.value
@@ -78,7 +82,8 @@ object TypelevelKernelPlugin extends AutoPlugin {
       onUnload.value.compose { (state: State) =>
         aliases.foldLeft(state) { (state, alias) => BasicCommands.removeAlias(state, alias) }
       }
-    }
+    },
+    tlPrePRSteps := List.empty
   )
 
   @nowarn("cat=deprecation")


zarthross:
Happy to add documentation if it looks like this code change will get accepted.

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
```diff
@@ -76,24 +76,32 @@ object TypelevelPlugin extends AutoPlugin {
...
 
-      val prePR = List("project /", "githubWorkflowGenerate") ++
+      List("githubWorkflowGenerate") ++
         List("+headerCreateAll").filter(_ => header) ++
         List("+scalafixAll").filter(_ => scalafix) ++
         List("+scalafmtAll", "scalafmtSbt").filter(_ => scalafmt) ++
-        List("javafmtAll").filter(_ => javafmt)
+        List("javafmtAll").filter(_ => javafmt) ++
+        List("compile", "test") ++
+        List("mimaReportBinaryIssues").filter(_ => mima)
```
armanbilge, line 92: See some discussion here.
- https://github.com/typelevel/sbt-typelevel/issues/204

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
```diff
@@ -76,24 +76,32 @@ object TypelevelPlugin extends AutoPlugin {
...
+    },
+    GlobalScope / tlCommandAliases ++= {
+      val header = tlCiHeaderCheck.value
+      val scalafmt = tlCiScalafmtCheck.value
+      val javafmt = tlCiJavafmtCheck.value
 
       val botHook = List("githubWorkflowGenerate") ++
         List("+headerCreateAll").filter(_ => header) ++
+        List("javafmtAll").filter(_ => javafmt) ++
         List("+scalafmtAll", "scalafmtSbt").filter(_ => scalafmt)
```
armanbilge, line 102: Steward only runs the bot hook only when there is an sbt-typelevel update PR. sbt-typelevel includes `sbt-scalafmt` but it doesn't include sbt-java-formatter. So there isn't really a situation where `javafmtAll` would need to be run to commit changes after an update. On the other hand, it's probably a harmless sink of Steward CPU cycles, esp. since few projects are using javafmt anyway.

kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
```diff
@@ -45,6 +45,9 @@ object TypelevelKernelPlugin extends AutoPlugin {
       "Command aliases defined for this build"
     )
 
+    lazy val tlPrePRSteps =
+      settingKey[List[String]]("Steps to be performed before a user submits a PR")
```
armanbilge, line 49: Mostly curious, why did you introduce this setting in the kernel plugin? I think we could introduce and set the default in the CI plugin, without need a blank default.

Also the doc can be more specific that these are the steps run by the `prePR` command alias.

zarthross: Added it to kernel because at my work we are using the Github plugin without using `TypelevelCiPlugin`, so for our use it made more sense to push it into Kernel.  Maybe there is an argument that we should be extending `TypelevelCiPlugin`... I`ll take a look next week, but  now that I`ve looked closer at the CIPlugin it might make sense.

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
zarthross: Happy to remove it, just felt weird without the symmetry....

armanbilge: Yeah, they are not really meant to be symmetrical. `prePR` is intended for human-made changes, `tlPrePrBotHook` is intended for bot-made changes. Anyway, I think it's fine to leave.

kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
armanbilge: If you are only using the kernel plugin, then it's fairly straightforward to do whatever you want directly with `tlCommandAliases`.

zarthross: While it is, and I did... I thought I would basically push back up the changes I was making for myself internally.  My thought process was that as people add different 'linters' or other things that are expected to pass in CI, such as scalafix/fmt, mima, or sbt-explicit-dependency, or sbt-missinglink , I wanted an easy way to aggregate those into a check at the the same time (in the same auto-plugin) as i was adding the github workflow. 

What ended up happening in my world is people started defining a command alias called 'build'  in each project and just put in the various steps, but somehow it would always be out of sync with what we actually did in CI.

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
zarthross: Gotcha, this wasn't clear in the existing docs on prePR but I'm happy to remove mima to go back into line with original intent.  Should I remove `compile` and `test` as well?  The docs seemed to indicate that at least `compile` was run?  https://typelevel.org/sbt-typelevel/faq.html#how-do-i-locally-prepare-my-pr-for-ci

> sbt-typelevel comes with a prePR command, which updates the GitHub workflow, generates headers, runs scalafmt, and clean compiles your code.

armanbilge: Oops, looks like the docs are not up to date. The original implementation did do a clean compile but we decided to change that in the linked issue.

kernel/src/main/scala/org/typelevel/sbt/TypelevelKernelPlugin.scala
zarthross: Alright, so the reason I'd prefer to keep prePR in kernel is because we have our own custom ci steps and we'd like the easier to extend prePR.

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
zarthross: Reverted it back to the original, removing mima and compile.

zarthross:
@armanbilge Thoughts on getting this merged?

armanbilge:
I'm pondering whether this belongs in kernel. Historically the mandate of kernel has been to implement public features that ideally would be upstreamed to sbt itself and private utilities that can be shared across the other plugins.

zarthross:
@armanbilge I can make a separate plugin in kernel project or a separate project?   In my opinion it doesn't belong in core either (we don't use core, we just use kernel, github, versioning, mima and scalafix)

armanbilge:
> In my opinion it doesn't belong in core either

I agree, I would say it belongs in the CI project—the notion of `prePR` is contingent on having some kind of repository with CI checks to merge into that repository.

zarthross:
😅 Completely understand your reasoning, but I was intentionally trying to avoid putting it in the ci project because we don't want to depend on it... We don't want the `TypelevelCiPlugin` autoplugin on our classpath so it doesn't accidentally activate.  We have our own Ci plugin we are implementing.

zarthross:
@armanbilge We could also change it so instead of using `prePR` we use some other separate command alias with a new name for this?  'build', 'prepare', 'lint', 'check'.. just to name some.

armanbilge:
> We could also change it so instead of using `prePR` we use some other separate command alias with a new name for this? 'build', 'prepare', 'lint', 'check'.. just to name some.

These are all interesting and IMO "opinionated" which I think we should try to avoid for the kernel plugin.

zarthross:
How about i make a new sbt project and move all the code in there?  Something independent of `prePR` and named differently, but serves the purpose of having an aggregated set of steps a user should perform before submitting a PR to minimize the failures seen in CI when making a PR?

armanbilge:
> but serves the purpose of having an aggregated set of steps a user should perform before submitting a PR to minimize the failures seen in CI when making a PR?

That's exactly the purpose of `prePR`, which is why it's named such :) and for it to auto-populate it makes sense to be defined in the plugin that configures the CI.

I appreciate your effort to upstream but it sounds like you are looking for something custom and maybe sbt-typelevel is a bit too opinionated for you? Why can't you depend on the sbt-typelevel-ci plugin?

Issue #735, opened 2024-08-11
Disable snapshots when publishing to central portal


armanbilge:
> The Central Publisher Portal does not support -SNAPSHOT releases, and deployments of -SNAPSHOT releases will have a version cannot be a SNAPSHOT error in their validation results. Versions should only be published to the Portal if they are intended to reach Maven Central.

https://central.sonatype.org/faq/snapshot-releases/

kubukoz:
"Disable snapshots", or "remove the -SNAPSHOT suffix"?

in other words, what do we mean by disabling?

armanbilge:
Good news, looks like they've added support for snapshots :)

https://central.sonatype.org/publish/publish-portal-snapshots/

PR #738, opened 2024-08-20
WIP: skip -SNAPSHOT suffix if publishing to new portal

kubukoz:
Just a draft for #735 to kickstart the discussion.

We don't have any scripted tests here, do we?

### DIFF
--- a/versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
+++ b/versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
@@ -25,6 +25,8 @@ import sbt._
 import sbt._
 
 import scala.util.Try
+import xerial.sbt.Sonatype
+import xerial.sbt.Sonatype.autoImport._
 
 import Keys._
 import Keys._
@@ -129,7 +131,10 @@ object TypelevelVersioningPlugin extends AutoPlugin {
         version += s"-$formatted"
       }
 
-      if (isSnapshot.value) version += "-SNAPSHOT"
+      val isPublishingToCentalPortal =
+        sonatypeCredentialHost.value == Sonatype.sonatypeCentralHost
+
+      if (isSnapshot.value && !isPublishingToCentalPortal) version += "-SNAPSHOT"
 
       version
     },


versioning/src/main/scala/org/typelevel/sbt/TypelevelVersioningPlugin.scala
```diff
@@ -129,7 +131,10 @@ object TypelevelVersioningPlugin extends AutoPlugin {
         version += s"-$formatted"
       }
 
-      if (isSnapshot.value) version += "-SNAPSHOT"
+      val isPublishingToCentalPortal =
+        sonatypeCredentialHost.value == Sonatype.sonatypeCentralHost
+
+      if (isSnapshot.value && !isPublishingToCentalPortal) version += "-SNAPSHOT"
```
armanbilge, line 137: Sonatype Central does not accept `-SNAPSHOT` versions.

> The Central Publisher Portal does not support -SNAPSHOT releases, and deployments of -SNAPSHOT releases will have a version cannot be a SNAPSHOT error in their validation results. Versions should only be published to the Portal if they are intended to reach Maven Central.

armanbilge: Oh sorry, I missed the `!` :)

Anyway, I feel it's disingenous for `isSnapshot` to be `true` but lacking the `-SNAPSHOT`.

armanbilge: Looking at how the setting is defined ...

https://github.com/sbt/sbt/blob/2f936295f158117baf9397ce8e9487bf9be8c87e/main/src/main/scala/sbt/Keys.scala#L520

armanbilge: Mmm, also disingenuous if `isSnapshot` is true but it's not published to the snapshot repo?

https://s01.oss.sonatype.org/content/repositories/snapshots/

armanbilge:
Here was my idea:
- https://github.com/typelevel/sbt-typelevel/pull/739

I guess the question I have is what does `isSnapshot` mean? Can something be a snapshot even if the version doesn't end in `-SNAPSHOT` and it's not published to a "snapshot" repository? Can you publish to Maven Central and call the artifacts snapshots?

armanbilge:
Pinging @majk-p @bpholt @j-mie6 who participated in the Discord discussion. Thoughts?

j-mie6:
My main thought is we should be careful about publishing anything that build tools may consider as a legitimate release. I could only imagine how annoying it would be to get warnings about updating to snapshots...

majk-p:
> I guess the question I have is what does `isSnapshot` mean? 

I have the same question. When using sbt-typelevel for my projects, with regards to publishing, I always valued that the setup works out of the box. By "works" I mean two main use cases:
1. When merged to main, CI pipeline produces a "snapshot" - something I can test, use where latest features are needed
2. When tagged on main, CI produces a proper version, such that will be picked by tools like Scala Steward

In that sense `isSnapshot` is something that's not picked up by Scala Steward.

Following that for a second, let's have a look at [`isSnapshot`](https://github.com/scala-steward-org/scala-steward/blob/main/modules/core/src/main/scala/org/scalasteward/core/data/Version.scala#L103) defined for Scala Steward:

```scala
  private def isSnapshot: Boolean =
    components.exists {
      case a: Version.Component.Alpha => a.isSnapshotIdent
      case _: Version.Component.Hash  => true
      case _                          => false
    }
```

It defines snapshot as anything that has a hash or is a special case of [`Alpha` release](https://github.com/scala-steward-org/scala-steward/blob/main/modules/core/src/main/scala/org/scalasteward/core/data/Version.scala#L167), so:

```scala
    final case class Alpha(value: String) extends Component {
      def isPreReleaseIdent: Boolean = order < 0
      def isSnapshotIdent: Boolean = order <= -7
      def order: Int =
        value.toUpperCase match {
          case "SNAP" | "SNAPSHOT" | "NIGHTLY" => -7
          case "FEAT" | "FEATURE"              => -6
          case "ALPHA" | "PREVIEW"             => -5
          case "BETA" | "B"                    => -4
          case "EA" /* early access */         => -3
          case "M" | "MILESTONE" | "AM"        => -2
          case "RC"                            => -1
          case _                               => 0
        }
    }
```

Back to sbt-typelevel, the current workaround with setting `tlUntaggedAreSnapshots := false` works fine because non-snapshots from mail look like: `0.1.2-1-4bce857` so they are snapshots as far as Steward is concerned.

IMO we have two main options for handling publishing to Central Portal
1. Agree that snapshots don't exist there and thus `tlUntaggedAreSnapshots := true` makes no sense. Instead set it to false as a default if central portal is configured. If user explicitly sets it to true, throw an error explaining what's going on
2. Overcome the fact that `-SNAPSHOT` suffix is not accepted by Central and in such case go with `-SNAP` or `-NIGHTLY`

I think I like the first option slightly better, but no hard opinion.

aartigao:
If it may be useful as a feedback, in my opinion, being a maintainer that came late to the party and is only able to publish to Central Portal, I must accept that the rules have changed and I have no option for publishing SNAPSHOTs anymore.

To me, publishing SNAPSHOT stuff in Maven Central is counterintuitive. At least in my case, if I want users to test something before final release, I'll just build an RC. So, from the options enumerated by @majk-p, I'd choose 👇🏽 

> Agree that snapshots don't exist there and thus tlUntaggedAreSnapshots := true makes no sense. Instead set it to false as a default if central portal is configured. If user explicitly sets it to true, throw an error explaining what's going on

armanbilge:
> 1. Agree that snapshots don't exist there and thus `tlUntaggedAreSnapshots := true` makes no sense.

SNAPSHOTS _do_ continue to exist, just not on Sonatype/Maven. For example, it's reasonable to `publishLocal` a `-SNAPSHOT` (I believe this is a concern for @j-mie6). So the setting `tlUntaggedAreSnapshots := true` can still be useful.

That's why in https://github.com/typelevel/sbt-typelevel/pull/739 instead of forcing it to `false`, we disable the CI publish workflow from running on untagged commits.

armanbilge:
Btw, thanks everyone for your thoughtful comments, I appreciate it :)

armanbilge:
@majk-p 

> That's something I'd like to avoid, since (if I get it right) eliminates the first use-case I mentioned above. Is that correct?

If `tlUntaggedAreSnapshots := true`, then we disable the CI publish workflow from running on untagged commits.

However, if `tlUntaggedAreSnapshots := false`, then we continue publishing every commit as usual. Is this the use-case you had in mind?

majk-p:
>  Is this the use-case you had in mind?

Yes, that's right. I guess my real questions should be what's the default `tlUntaggedAreSnapshots` for new projects (using central). Asked the other way around, is sbt-typelevel publishing untagged artifacts out of the box for new projects?

armanbilge:
> Asked the other way around, is sbt-typelevel publishing untagged artifacts out of the box for new projects?

It's not. My concern is that it can confuse tooling so it has to be opt-in. FS2 used to publish like this and I remember having some problems. Still, it looks like Steward and Scaladex are able to handle it okay so maybe it's less of a problem now?

Screenshot 2024-08-21 at 2 24 31 PM

armanbilge:
I think I want to move ahead with https://github.com/typelevel/sbt-typelevel/pull/739 but I'm curious if there are any more thoughts.

kubukoz:
Like they say, the best way to find out the answer is sometimes to give the wrong answer first :D

PR #739, opened 2024-08-20
Disable CI snapshots if publishing to Central host

armanbilge:
Closes https://github.com/typelevel/sbt-typelevel/issues/735. Alternative to https://github.com/typelevel/sbt-typelevel/pull/738.

### DIFF
--- a/core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
+++ b/core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
@@ -20,6 +20,7 @@ import de.heikoseeberger.sbtheader.HeaderPlugin
 import org.typelevel.sbt.gha.GenerativePlugin
 import org.typelevel.sbt.gha.GitHubActionsPlugin
 import sbt._
+import xerial.sbt.Sonatype
 
 import Keys._
 import TypelevelKernelPlugin.autoImport._
@@ -48,8 +49,10 @@ object TypelevelPlugin extends AutoPlugin {
   import TypelevelCiPlugin.autoImport._
   import TypelevelSettingsPlugin.autoImport._
   import TypelevelSonatypeCiReleasePlugin.autoImport._
+  import TypelevelVersioningPlugin.autoImport._
   import GenerativePlugin.autoImport._
   import GitHubActionsPlugin.autoImport._
+  import Sonatype.autoImport._
 
   override def buildSettings = Seq(
     organization := "org.typelevel",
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
```diff
@@ -64,7 +67,14 @@ object TypelevelPlugin extends AutoPlugin {
     licenses += License.Apache2,
     tlCiHeaderCheck := true,
     tlCiScalafmtCheck := true,
-    tlCiReleaseBranches := Seq("main"),
+    tlCiReleaseBranches := {
+      val central =
+        sonatypeCredentialHost.value.equalsIgnoreCase(Sonatype.sonatypeCentralHost)
+      if (central && tlUntaggedAreSnapshots.value)
+        Seq() // Sonatype Central does not support snapshots
```
majk-p, line 74: Do you mind adding a warning log that user should change their sbt settings since this combination is not supported? Perhaps linking to sonatype central docs for reference?

armanbilge: @majk-p the tricky thing is that this combination _is_ supported and a user doesn't have to change their sbt settings. It's totally valid for the user to want `-SNAPSHOT` versions and not publish them in CI.

The only combination that's unsupported is attempting to publish `-SNAPSHOT`s to the new central post.

armanbilge: > The only combination that's unsupported is attempting to publish `-SNAPSHOT`s to the new central post.

Let's add a warning lot for that, is that what you meant?

majk-p: Yes, that's what I meant, sorry. Perhaps if saying it's unsupported, the log could sat the behavior is likely unexpected and no snapshot will be published to central

bpholt:
I think this approach seems wise! Thanks for working through it.

core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala
armanbilge: > Do you mind adding a warning log that user should change their sbt settings since this combination is not supported?

Actually, it's not clear to me where to put the warning 🤔 

We don't want to put the warning in `tlCiReleaseBranches` because there's nothing invalid happening there. It's only if the user overrides its value that we may need to issue a warning, but I don't know how to hook into that esp. when loading the build. I think I might just make this a follow-up issue.

PR #751, opened 2024-10-05
Update sbt-scalafix to 0.13.0

alexcardell:
0.13.0 allows using `OrganizeImports.removeUnused` in Scala 3.3.4, which is very helpful

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ name := "sbt-typelevel"
 
 import org.typelevel.sbt.gha.{PermissionScope, PermissionValue, Permissions}
 
-ThisBuild / tlBaseVersion := "0.7"
+ThisBuild / tlBaseVersion := "0.8"
 ThisBuild / crossScalaVersions := Seq("2.12.20")
 ThisBuild / developers ++= List(
   tlGitHubDev("armanbilge", "Arman Bilge"),
--- a/scalafix/build.sbt
+++ b/scalafix/build.sbt
@@ -1,1 +1,1 @@
-addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.12.1")
+addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.13.0")


armanbilge:
Dupe of https://github.com/typelevel/sbt-typelevel/pull/749, although we didn't bump the base version 🤔

alexcardell:
Oh my bad, I only checked the latest release, doh.

I bumped because I assumed it would be incompatible for any plugins depending on the scalafix plugin

armanbilge:
> I bumped because I assumed it would be incompatible for any plugins depending on the scalafix plugin

Yes, I think you did the right thing. I am going to think more about it.

mzuehlke:
Bumping the version because one of our dependency did a change is not an easy question.
I often argued that a library tool is only responsible for the their own API. If I as a client use `sbt-typelevel` and rely on the functionality on one of the dependencies of `sbt-typelevel` than I need to make sure I add this to my own. In other words never rely on transitive dependencies.

When we would follow this argumentation we won't have to bump our base version.

armanbilge:
@mzuehlke thanks for explaining your reasoning.

> When we would follow this argumentation we won't have to bump our base version.

When I follow your argument (and I think it's a reasonable one), I come to the conclusion that we should not bump the Scalafix dependency.

> If I as a client use `sbt-typelevel` and rely on the functionality on one of the dependencies of `sbt-typelevel` than I need to make sure I add this to my own. In other words never rely on transitive dependencies.

Yes, I can agree with this. If you want the latest Scalafix features, then you can directly add the Scalafix dependency to your build. This is opt-in.

The problem is that when we ship the latest (breaking) Scalafix, then we _force_ the breakage onto users without an ability for them to opt-out.

So I think your argument about users/clients taking responsibility for transitive dependencies, only makes sense if we don't force breaking versions onto them.

---

However ... there is also an argument that sbt-typelevel should make builds easy without needing to add all the dependencies (similar to the scala-cli philosophy).

---

Thoughts? Thanks for the discussion.

---

I was about to post, but I had an idea for how I want to move forward.

We'll publish the Scalafix bump without bumping the sbt-typelevel base version. If this causes breakage for a user, they can choose to pin `sbt-typelevel-scalafix` to v0.7.3 while continuing to use the latest `sbt-typelevel` until they are able to complete the Scalafix upgrade.

PR #756, opened 2024-10-10
Use `actions/setup-java` to install GraalVM

mzuehlke:
Since v4.4.0 graalvm can be installed using setup-java https://github.com/actions/setup-java/releases/tag/v4.4.0 (for legacy versions setup-graalvm is still needed)

Not sure if we need to expose these additional distributions: 
- https://github.com/actions/setup-java/releases/tag/v4.3.0 supports SapMachine 
- https://github.com/actions/setup-java/releases/tag/v3.13.0 supports Dragonwell


armanbilge (approved):
nice! thanks

PR #761, opened 2024-10-14
Trying a shorter job name for better display in GHA

mzuehlke:
Before
Bildschirmfoto 2024-10-14 um 16 05 46

After:
Bildschirmfoto 2024-10-14 um 16 06 02

**Feedback welcome.**
 I really dislike this part of the GH UI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ concurrency:
 
 jobs:
   build:
-    name: CI
+    name: Test
     strategy:
       matrix:
         os: [ubuntu-22.04, macos-latest, windows-latest]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ concurrency:
 
 jobs:
   build:
-    name: Build and Test
+    name: CI
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -12,15 +12,15 @@ pull_request_rules:
   - or:
     - body~=labels:.*early-semver-patch
     - body~=labels:.*early-semver-minor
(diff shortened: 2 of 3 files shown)


armanbilge:
I don't have a strong preference ... this is just something we've inherited from sbt-github-actions. My thought is CI refers to the entire workflow (build, test, docs, publish, etc) vs build-and-test is a specific job that is part of CI.

> I really dislike this part of the GH UI

Yep, I have struggled many times to figure out which cell of the matrix a failure happened in ...

valencik:
This is definitely an improvement in my opinion.
We get to see the Scala version and most of the JVM name.

If we could ditch the `-latest` suffix from the OS names, that would probably let you see everything of importance.

mzuehlke:
With the switch from `ubuntu-latest` to `ubuntu-22.04` I am bit reluctant to remove OS version.
"CI" was just my first idea, mostly because it is short 😉 
"Test" would be another suggestion (instead of "Build and Test"

mzuehlke:
Instead of leaving the generation of the `job.name` to the automation we could generate a name, but only list axes from the matrix that have more than one dimension. In this project that would remove the `scala` version 8always "2.12" and the `project` (always "sbt-typelevelJVM").

armanbilge (approved):
Thanks! Let's try this :)

bpholt:
FWIW, I like the new name, but the change has caused me a fair amount of admin work updating branch protection rules in all the repos where we're using sbt-typelevel. Since we have required status checks in place, no PRs can be merged following this change without an admin manually logging in an updating the rules to reflect the new names. Had I realized this change was coming, I would have 👎 solely due to the amount of downstream work involved.

(I'm also worried about the change from `ubuntu-latest` to `ubuntu-22.04`. When that Ubuntu version is bumped, it's going to trigger another round of the same administrative headache. 😩)

If anyone is aware of a way to automate the changes to the required status checks, I would be very grateful!

mzuehlke:
@bpholt it is probably too late fro that, but an option could be to make this configurable.
Then you could have avoided this tedious admin work 😢

PR #762, opened 2024-10-16
pin to ubuntu-22.04 as default runner OS

samspills:
This week we saw a lot of chaos because the `ubuntu-latest` runner tag was switched to the `ubuntu-24.04` image, away from the `ubuntu-22.04` image[^1]. 

I'm proposing that we pin the ubuntu runner OS to 22.04 as the default value so we can avoid that chaos in the future. By pinning we can
- test here before updating to a new runner tag, preventing CI for sbt-typelevel users from breaking unexpectedly (breaking due to this kind of change at least lol)
- communicate with users via a major release + release notes if a major change like this happens again

[^1]: And that switch is now being [rolled back](https://github.com/actions/runner-images/issues/10636#issuecomment-2417293457) for the time being 😵

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
     name: Build and Test
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-22.04, macos-latest, windows-latest]
         scala: [2.12]
         java:
           - temurin@8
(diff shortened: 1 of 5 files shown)


valencik:
Doing this would also save us an extra character width in the job names!
re: https://github.com/typelevel/sbt-typelevel/pull/761

armanbilge:
👍 to pinning. Why go back to 22.04 / when _should_ we go to 24.04 given that things are working now (I think 😅). I guess we can wait until GH makes 24.04 latest again, as a sign of stability? idk.

samspills:
> Why go back to 22.04 / when should we go to 24.04 given that things are working now (I think 😅). I guess we can wait until GH makes 24.04 latest again, as a sign of stability? idk.

I had the same thinking/question. I chose 22.04 so it would match the _current_ behaviour of `latest` just in case, but it's not a strongly held opinion

armanbilge:
Follow-up idea
- https://github.com/typelevel/sbt-typelevel/issues/764

PR #810, opened 2025-04-06
Tweak the publication condition to prevent publishing from forked repos

danicheg:
One might argue that users can use `githubWorkflowPublishCond` to prevent publishing from forked repos. Sure, but I’m confident this condition should be the default. Adding it manually to the build of dozens of projects isn’t exactly ideal, either.
This is basically the same as what we did in #720, but adding the same SBT setting again just feels a bit extra to me 🤷🏻

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,7 @@ tags
 .bloop/
 metals.sbt
 .vscode
+
+# Macos
+.DS_Store
+.Trashes
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,3 @@ tags
 .bloop/
 metals.sbt
 .vscode
-
-# Macos
-.DS_Store
-.Trashes
--- a/ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
+++ b/ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
@@ -23,6 +23,7 @@ import org.typelevel.sbt.gha.GitHubActionsPlugin
 import org.typelevel.sbt.gha.WorkflowStep
 import sbt._
 
+import scala.annotation.nowarn
 import scala.language.experimental.macros
 
 object TypelevelCiPlugin extends AutoPlugin {
@@ -52,6 +53,7 @@ object TypelevelCiPlugin extends AutoPlugin {
     lazy val tlCiStewardValidateConfig = settingKey[Option[File]](
       "The location of the Scala Steward config to validate (default: `.scala-steward.conf`, if exists)")
 
+    @deprecated("Use `githubWorkflowForkCondition` instead", "0.7.8")
     lazy val tlCiForkCondition =
       settingKey[String](
         "Condition for checking on CI whether this project is a fork of another (default: `github.event.repository.fork == false`)")
@@ -60,6 +62,7 @@ object TypelevelCiPlugin extends AutoPlugin {
 
   import autoImport._
 
+  @nowarn("cat=deprecation")
   override def buildSettings = Seq(
     tlCiHeaderCheck := false,
     tlCiScalafmtCheck := false,
@@ -68,7 +71,7 @@ object TypelevelCiPlugin extends AutoPlugin {
     tlCiMimaBinaryIssueCheck := false,
     tlCiDocCheck := false,
     tlCiDependencyGraphJob := true,
-    tlCiForkCondition := "github.event.repository.fork == false",
+    tlCiForkCondition := githubWorkflowForkCondition.value,
     githubWorkflowTargetBranches ++= Seq(
       "!update/**", // ignore steward branches
       "!pr/**" // escape-hatch to disable ci on a branch
@@ -145,7 +148,7 @@ object TypelevelCiPlugin extends AutoPlugin {
     githubWorkflowJavaVersions := Seq(JavaSpec.temurin("8")),
     githubWorkflowAddedJobs ++= {
       val ghEventCond = "github.event_name != 'pull_request'"
-      val jobCond = s"${tlCiForkCondition.value} && $ghEventCond"
+      val jobCond = s"${githubWorkflowForkCondition.value} && $ghEventCond"
 
       val dependencySubmission =
         if (tlCiDependencyGraphJob.value)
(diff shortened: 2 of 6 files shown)


danicheg:
I was just thinking again about adding a dedicated SBT setting for the fork condition. Maybe introducing `githubWorkflowForkCondition` with a default value of `github.event.repository.fork == false` isn't such a bad idea — especially if we deprecate the `tlCiForkCondition` and substitute its default with the newly added `githubWorkflowForkCondition`. WDYT?

bpholt:
I think it is a good idea not to publish from forks by default, but I can envision situations where users legitimately would want to do so, so I don't think it should be disabled without a way to re-enable it. A dedicated sbt setting seems reasonable to me.

.gitignore
```diff
@@ -17,3 +17,7 @@ tags
 .bloop/
 metals.sbt
 .vscode
+
+# Macos
+.DS_Store
+.Trashes
```
armanbilge, line 23: Since these are specific to your operating system (and not this project) they should go in your global `.gitignore` configuration

ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
```diff
@@ -145,7 +148,7 @@ object TypelevelCiPlugin extends AutoPlugin {
     githubWorkflowJavaVersions := Seq(JavaSpec.temurin("8")),
     githubWorkflowAddedJobs ++= {
       val ghEventCond = "github.event_name != 'pull_request'"
-      val jobCond = s"${tlCiForkCondition.value} && $ghEventCond"
+      val jobCond = s"${githubWorkflowForkCondition.value} && $ghEventCond"
```
armanbilge, line 151: Unfortunately I think we need to keep this as `tlCiForkCondition` in case a user has overridden it, otherwise this change will not be compatible.

docs/gha.md
```diff
@@ -80,6 +80,7 @@ Any and all settings which affect the behavior of the generative plugin should b
 - `githubWorkflowEnv` : `Map[String, String]` – An environment which is global to the entire **ci.yml** workflow. Defaults to `Map("GITHUB_TOKEN" -> "${{ secrets.GITHUB_TOKEN }}")` since it's so commonly needed.
 - `githubWorkflowAddedJobs` : `Seq[WorkflowJob]` – A convenience mechanism for adding extra custom jobs to the **ci.yml** workflow (though you can also do this by modifying `githubWorkflowGeneratedCI`). Defaults to empty.
 - `githubWorkflowConcurrency` : `Option[Concurrency]` - Use concurrency to ensure that only a single workflow within the same concurrency group will run at a time. Defaults to `${{ github.workflow }} @ ${{ github.ref }}`.
+- `githubWorkflowForkCondition` : `String` — A condition in the workflow to determine if this project is a fork of another (default: `github.event.repository.fork == false`)
```
armanbilge, line 83:
```suggestion
- `githubWorkflowForkCondition` : `String` - A condition to determine if this workflow is running on a fork (default: `github.event.repository.fork == false`)
```

github-actions/src/main/scala/org/typelevel/sbt/gha/GenerativeKeys.scala
```diff
@@ -113,6 +113,9 @@ trait GenerativeKeys {
     s"Permissions to use for the global workflow (default: None)")
   lazy val githubWorkflowAddedJobs = settingKey[Seq[WorkflowJob]](
     "A list of additional jobs to add to the CI workflow (default: [])")
+  lazy val githubWorkflowForkCondition =
+    settingKey[String](
+      "A condition in the workflow to determine if this project is a fork of another (default: `github.event.repository.fork == false`)")
```
armanbilge, line 118:
```suggestion
      "A condition to determine if this workflow is running on a fork (default: `github.event.repository.fork == false`)")
```

armanbilge (commented):
Thank you, this is a really good change. Sorry it took me so long to review. I have just a couple nitpicks.

ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
danicheg: Hmm, yes, backward compatibility is a bit affected here, but does it really make sense to have `githubWorkflowForkCondition` and `tlCiForkCondition` in an unaligned state? I couldn’t find any projects on GitHub that use `tlCiForkCondition`, so maybe we could make an announcement about the breaking change and accept it? I generally prefer maintaining backward compatibility, but in this case, users who have overridden `tlCiForkCondition` will just get a failed build — since `githubWorkflowCheck` will fail (they haven’t set `githubWorkflowForkCondition`, and `tlCiForkCondition` is overridden) — so the migration should be fairly straightforward...
Let me know if you're strongly against it. If so, I’ll have no choice but to comply.

armanbilge: > Hmm, yes, backward compatibility is a bit affected here, but does it really make sense to have `githubWorkflowForkCondition` and `tlCiForkCondition` in an unaligned state?

I agree, It does not make sense for them to be unaligned, and the deprecation will nudge users towards `githubWorkflowForkCondition`.

> but in this case, users who have overridden `tlCiForkCondition` will just get a failed build — since `githubWorkflowCheck` will fail (they haven’t set `githubWorkflowForkCondition`, and `tlCiForkCondition` is overridden)

Actually, they won't get a failed build: Scala Steward will automatically auto-generate the workflow, so the build passes. So it would be easy to merge an update that actually changes the behavior.

> so maybe we could make an announcement about the breaking change and accept it

A breaking release of sbt-typelevel is on the horizon anyway, because of the Sonatype Central changes. When that happens, we can just get rid of the deprecated `tlCiForkCondition`.

armanbilge (commented):
I'm so sorry—I just realize one more important bikeshed I think.

Currently, `githubWorkflowForkCondition == true` if the repository is _not_ a fork, which is confusing. (This was the original mistake of `tlCiForkCondition`).

So either we should call it `githubWorkflowNotForkCondition` _or_ change the definition to test `== true`, and then adjust the usecases.

PR #811, opened 2025-04-17
Add continueOnError option to WorkflowStep

hugo-vrijswijk:


github-actions/src/main/scala/org/typelevel/sbt/gha/GenerativePlugin.scala
```diff
@@ -348,8 +355,9 @@ ${indent(rendered.mkString("\n"), 1)}"""
       commands: List[String],
       params: Map[String, String],
       renderedShell: String,
-      renderedWorkingDirectory: String) =
-    renderedShell + renderedWorkingDirectory + "run: " + wrap(
+      renderedWorkingDirectory: String,
+      renderedContinueOnError: String) =
```
armanbilge, line 359: To fix bincompat, add a deprecated overload of this method with the old signature that passes an empty string to `renderedContinueOnError`

github-actions/src/main/scala/org/typelevel/sbt/gha/WorkflowStep.scala
```diff
@@ -144,8 +146,18 @@ object WorkflowStep {
         env: Map[String, String] = Map(),
         params: Map[String, String] = Map(),
         timeoutMinutes: Option[Int] = None,
-        workingDirectory: Option[String] = None): Run =
-      Impl(commands, id, name, cond, env, params, timeoutMinutes, workingDirectory)
+        workingDirectory: Option[String] = None,
+        continueOnError: Boolean = false): Run =
```
armanbilge, line 150: Same thing here: unfortunately, adding a default parameter (`= false`) will not fix bincompat. However, adding a deprecated overload with the old signature, that passes `continueOnError = false` to the new overload will work.

armanbilge (changes requested):
Thanks for the PR! Sorry it took me a while to circle back for review.

The good news is that it is possible to make these changes bincompatibly. You can't change existing method signatures, but you can deprecate them and add new method signatures.

Sometimes, MiMa doesn't recognize internal private implementations, so you may need to add filters for those.

PR #818, opened 2025-05-13
Change sort order of versions so final releases are last not first.

zarthross:
There was a minor issue where if a commit was tagged with both an RC and a final release, "v0.6.0-RC1" and "v0.6.0", the current version would be the most recent RC and not the final version.  This PR switches the bias on sort, so prereleases come before final releases.

### DIFF
--- a/kernel/src/main/scala/org/typelevel/sbt/kernel/V.scala
+++ b/kernel/src/main/scala/org/typelevel/sbt/kernel/V.scala
@@ -51,8 +51,8 @@ private[sbt] final case class V(
         if (z != 0) return z
         (this.prerelease, that.prerelease) match {
           case (None, None) => 0
-          case (Some(_), None) => 1
-          case (None, Some(_)) => -1
+          case (Some(_), None) => -1
+          case (None, Some(_)) => 1
           case (Some(thisPrerelease), Some(thatPrerelease)) =>
             // TODO not great, but not everyone uses Ms and RCs
             thisPrerelease.compare(thatPrerelease)
--- a/kernel/src/test/scala/org/typelevel/sbt/kernel/VSuite.scala
+++ b/kernel/src/test/scala/org/typelevel/sbt/kernel/VSuite.scala
@@ -17,6 +17,7 @@
 package org.typelevel.sbt.kernel
 
 import munit.FunSuite
+import scala.util.Random
 
 class VSuite extends FunSuite {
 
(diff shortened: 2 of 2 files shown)


zarthross:
Note: https://github.com/typelevel/governance/issues/163#issuecomment-2877608060

samspills (approved):

mzuehlke (approved):

kernel/src/main/scala/org/typelevel/sbt/kernel/V.scala
```diff
@@ -51,8 +51,8 @@ private[sbt] final case class V(
         if (z != 0) return z
         (this.prerelease, that.prerelease) match {
           case (None, None) => 0
-          case (Some(_), None) => 1
-          case (None, Some(_)) => -1
+          case (Some(_), None) => -1
+          case (None, Some(_)) => 1
           case (Some(thisPrerelease), Some(thatPrerelease)) =>
             // TODO not great, but not everyone uses Ms and RCs
```
rossabaker, line 57: This has a ton of corner cases.  Which corners (e.g., where to look for numbers, "well-known qualifiers") depends on which spec (e.g., SemVer, Coursier, Ivy, Maven).  But this PR will make it do the right thing a significantly higher percentage of the time in practice in all of them.

rossabaker (approved):

kernel/src/test/scala/org/typelevel/sbt/kernel/VSuite.scala
```diff
@@ -178,4 +178,12 @@ class VSuite extends FunSuite {
     assertEquals(pre1 < nopre1, true)
   }
 
+  test("pre-release is before release") {
+    val release = V(0, 1, Some(1), None)
+    val rc1 = V(0, 1, Some(1), Some("RC1"))
+    val rc2 = V(0, 1, Some(1), Some("RC2"))
+    val expected = List(rc1, rc2, release)
```
armanbilge, line 185: can we add an M1/M2 to this test as well?

armanbilge (commented):
Thanks, good catch! There was actually supposed to be an existing test for this, but the test was written incorrectly unfortunately. We can fix it:

https://github.com/typelevel/sbt-typelevel/blob/2b348886674f07751d76298376e10027356d280c/kernel/src/test/scala/org/typelevel/sbt/kernel/VSuite.scala#L172-L179

kernel/src/test/scala/org/typelevel/sbt/kernel/VSuite.scala
zarthross: Done, also merged this test with the previous one and fixed it.

armanbilge (approved):

PR #835, opened 2025-09-12
Update sbt-header to 5.11.0

mzuehlke:
closes: #833

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ name := "sbt-typelevel"
 
 import org.typelevel.sbt.gha.{PermissionScope, PermissionValue, Permissions}
 
-ThisBuild / tlBaseVersion := "0.8"
+ThisBuild / tlBaseVersion := "0.9"
 ThisBuild / crossScalaVersions := Seq("2.12.20")
 ThisBuild / developers ++= List(
   tlGitHubDev("armanbilge", "Arman Bilge"),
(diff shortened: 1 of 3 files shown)


rossabaker (approved):
This is going to need to be a major version bump of this plugin, I think.

armanbilge (changes requested):
Indeed. Should bump the base version appropriately in this PR. Any compelling reasons to make this update?

rossabaker:
This update will be necessary for eventual SBT 2 support, I think?  I don't think it's urgent in the short term, but needed in the long term.

I'm also contemplating an improvement to sbt-header's SPDX handling.  That would be a subsequent release.

mzuehlke:
As @rossabaker sad, we will need a newer version latest when we want to support sbt 2.
In the meantime it depends if we want to always include updates to all used sbt plugins.
Which would lead to more often base version bumps. Or do this very selectively.

satorg:
Just checknig – is there anything missing in the pull request that is blocking it from getting merged?

satorg:
@mzuehlke , could you update your branch and resolve conflicts please?
It looks like your PR is otherwise pretty much ready to go to `main`.
Thank you!

guillaumebadin:
Hi 👋 — just a note that this is blocking downstream projects. We maintain [edomata](https://github.com/beyond-scale-group/edomata) which depends on sbt-typelevel, and Scala Steward opened a PR to update sbt-header to 5.11.0 ([beyond-scale-group/edomata#21](https://github.com/beyond-scale-group/edomata/pull/21)) which fails with the `headerLicenseStyle` / `headerMappings` / `headerLicense` type conflict between the old and new package names.

Would love to see this merged when the time is right. Thanks for the work on it!

armanbilge:
@guillaumebadin thanks for bringing that to my attention!

> just a note that this is blocking downstream projects

For my own understanding, are there important changes in this new version of sbt-header that you are being blocked from using?

So far, my understanding is that the release is mostly/entirely about adding support for sbt 2. We will have to do that work in sbt-typelevel as well.

PR #845, opened 2025-10-24
Allow defining outputs for GHA jobs

rpiaggio:
Allow defining `outputs` for jobs: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idoutputs.

Values are automatically interpreted as expressions, ie, wrapped in `${{ ... }`.

### DIFF
--- a/github-actions/src/main/scala/org/typelevel/sbt/gha/WorkflowJob.scala
+++ b/github-actions/src/main/scala/org/typelevel/sbt/gha/WorkflowJob.scala
@@ -114,7 +114,7 @@ object WorkflowJob {
       outputs
     )
 
-  // Kept for binary compatibility. It can be removed in a future major release.
+  @deprecated("Kept for binary-compatibility", "0.8.4")
   def apply(
       id: String,
       name: String,


build.sbt
```diff
@@ -3,7 +3,7 @@ name := "sbt-typelevel"
 import org.typelevel.sbt.gha.{PermissionScope, PermissionValue, Permissions}
 import com.typesafe.tools.mima.core._
 
-ThisBuild / tlBaseVersion := "0.8"
+ThisBuild / tlBaseVersion := "0.9"
```
bpholt: Can this be accomplished without bumping the version here? This is effectively a major version change, with all that entails for libraries. I haven't looked closely at whether it's necessary or not.

github-actions/src/main/scala/org/typelevel/sbt/gha/WorkflowJob.scala
```diff
@@ -86,6 +88,7 @@ object WorkflowJob {
       container: Option[JobContainer] = None,
       environment: Option[JobEnvironment] = None,
       concurrency: Option[Concurrency] = None,
+      outputs: Map[String, String] = Map(),
```
armanbilge: This is the binary-breaking change. To restore binary-compatibility, we need to keep an overload of `apply` with the old signature, that can delegate to the new one.

It's also a source-breaking change: `outputs` should be added as the final parameter in the list.

github-actions/src/main/scala/org/typelevel/sbt/gha/WorkflowJob.scala
```diff
@@ -107,7 +112,54 @@ object WorkflowJob {
...
+      matrixExcs,
+      runsOnExtraLabels,
+      container,
+      environment,
+      concurrency,
+      timeoutMinutes,
+      Map()
     )
 
   private final case class Impl(
```
armanbilge: We don't need to preserve the binary-compatibility of `Impl` because it is a private implementation class. So most of the changes below are unnecessary. You may need to add MiMa filters.

github-actions/src/main/scala/org/typelevel/sbt/gha/GenerativePlugin.scala
```diff
@@ -210,6 +210,21 @@ object GenerativePlugin extends AutoPlugin {
 ${indent(rendered.mkString("\n"), 1)}"""
     }
 
+  def compileOutputs(outputs: Map[String, String]): String =
```
armanbilge: To dogfood this, can we add a couple steps to the sbt-typelevel build that exercise this feature?

rpiaggio: Ok, I've implemented a couple of jobs that produce and consume `outputs`, the latter failing if the output is missing.

PR #854, opened 2025-12-07
Update `actions/checkout` to v6

xuwei-k:
https://github.com/actions/checkout/releases/tag/v6.0.0


rossabaker (commented):
I was going to say this requires the runner from July, but I guess v5 already did.  I haven't run into any issues using v6 on public infrastructure.  It's probably fine?

PR #861, opened 2026-01-09
Change concurrency class to handle more complex expressions for cancellation.

zarthross:
Github actions allows you to set cancellation based on an expression.
Changing the `cancelInProgress` to a string allows me to set the
cancellation based on branch which would help my team prevent
cancellation of main branch builds in the middle of an automatic publish
leaving us in a half published state.

Example: 
[docs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#example-only-cancel-in-progress-jobs-on-specific-branches)

``` yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ !contains(github.ref, 'release/')}}
```


zarthross:
@armanbilge ?

armanbilge (approved):
Thanks for that follow-up. Nicely done!

PR #874, opened 2026-02-25
Add opt-in support for TASTy-MiMa to check Scala 3 compatibility in CI

### LINKED ISSUE #547: Support TASTy-MiMa
Probably start opt-in, and then go from there.

https://github.com/scalacenter/tasty-mima

One difference from ordinary MiMa is I think we only need to check against the last version? i.e. not all previous versions 🤔

the-ivii:
Fixes #547 

i have added opt-in support for sbt-tasty-mima. now, Scala 3 projects can check for TASTy compatibility directly in CI. based on discussion in the issue, it currently defaults to false. it only checks against the latest previous version instead of the whole release history

### DIFF
--- a/mima/src/main/scala/org/typelevel/sbt/TypelevelMimaPlugin.scala
+++ b/mima/src/main/scala/org/typelevel/sbt/TypelevelMimaPlugin.scala
@@ -17,16 +17,18 @@
 package org.typelevel.sbt
 
 import com.typesafe.tools.mima.plugin.MimaPlugin
+import sbttastymima.TastyMiMaPlugin
 import org.typelevel.sbt.kernel.GitHelper
 import org.typelevel.sbt.kernel.V
 import sbt._
 
 import Keys._
 import MimaPlugin.autoImport._
+import TastyMiMaPlugin.autoImport._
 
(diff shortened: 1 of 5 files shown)


mima/src/main/scala/org/typelevel/sbt/TypelevelMimaPlugin.scala
```diff
@@ -36,6 +38,7 @@ object TypelevelMimaPlugin extends AutoPlugin {
         "A map scalaBinaryVersion -> version e.g. Map('2.13' -> '1.5.2', '3' -> '1.7.1') used to indicate that a particular crossScalaVersions value was introduced in a given version (default: empty).")
     lazy val tlMimaPreviousVersions = settingKey[Set[String]](
       "A set of previous versions to compare binary-compatibility against (default: auto-populated from git tags and the tlVersionIntroduced setting)")
+    lazy val tlTastyMima = settingKey[Boolean]("Enable TASTy-MiMa (default: false)")
```
armanbilge: What is the purpose of this setting?

armanbilge (commented):
Thanks for the PR. How are Scala 2 artifacts handled?

PR #881, opened 2026-03-13
Skip Scala.js sourcemap mapping when there are uncommitted changes

hugo-vrijswijk:
Skips the `mapSourceURI` flag when there are uncommitted changes, since it will be very unlikely to map to a remote source.
Source maps can still not exist when the commit is not pushed, but this will at least prevent the source map from being completely broken during local development.


armanbilge (approved):
Thanks, good fix.

An alternative implementation I considered is if we should only set the GH sourcemaps when running in GHA CI. But that would be problematic in the unusual but not impossible event that a project has to be published locally. So I think this is the right compromise.

PR #882, opened 2026-04-08
Enable dependency-submission on same-repo pull requests

antoniojimeneznieto:
The dependency-submission job currently only runs on push, so vulnerabilities are only detected after merge. This PR enables it for same-repo PRs too, so projects can catch vulnerable dependencies before merging.

Fork PRs are still excluded because their read-only token can't call the Submission API.

Refs: #340, #342

### DIFF
--- a/ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
+++ b/ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
@@ -20,6 +20,8 @@ import org.typelevel.sbt.NoPublishGlobalPlugin.noPublishModulesIgnore
 import org.typelevel.sbt.gha.GenerativePlugin
 import org.typelevel.sbt.gha.GenerativePlugin.autoImport._
 import org.typelevel.sbt.gha.GitHubActionsPlugin
+import org.typelevel.sbt.gha.PermissionValue
+import org.typelevel.sbt.gha.Permissions
 import org.typelevel.sbt.gha.WorkflowStep
 import sbt._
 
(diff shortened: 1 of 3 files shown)


ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
```diff
@@ -144,8 +151,14 @@ object TypelevelCiPlugin extends AutoPlugin {
     },
     githubWorkflowJavaVersions := Seq(JavaSpec.temurin("8")),
     githubWorkflowAddedJobs ++= {
-      val ghEventCond = "github.event_name != 'pull_request'"
-      val jobCond = s"${tlCiForkCondition.value} && $ghEventCond"
+      val sameRepoPrCond =
+        "github.event.pull_request.head.repo.full_name == github.repository"
```
armanbilge, line 155: Could we use `tlCiForkCondition` for this, or is it checking something different? 🤔

armanbilge: Oh sorry, dumb question, i see it's something different 🤔

antoniojimeneznieto: If I've got all of this right (which might not be the case 😅), the idea is to run the vulnerability detection job on PRs from the same repo so Scala Steward or a maintainer's branch which works because those PRs run with a token that can be granted `contents: write`.

However, it won't work for PRs from a contributor's fork, because the Submission API requires `contents: write`, and PRs from forks always run with a read-only token.

The `sameRepoPrCond` condition is checking this.

armanbilge: Sorry, I'm still confused 😅 

> the idea is to run the vulnerability detection job on PRs from the same repo

> However, it won't work for PRs from a contributor's fork

That makes sense. Isn't this equivalent to checking if CI is running on a fork? Which is what `tlCiForkCondition` represents.

antoniojimeneznieto: I think that tlCiForkCondition doesn't cover the case where a fork PR runs on the original repo's CI 🤔

armanbilge: Oh I see, I think you are right! PR runs CI in the original repo, but without the write permissions. So confusing 😅 thanks for clarifying :)

PR #889, opened 2026-05-07
Add `tlCiLintCommands` key

### LINKED ISSUE #888: Public hook for appending commands to the "Check headers and formatting" lint step
## Problem

sbt-typelevel's generated `Test` job emits a step named `"Check headers and formatting"` that runs sbt commands like:

```sh
sbt '++ ${{ matrix.scala }}' headerCheckAll scalafmtCheckAll 'project /' scalafmtSbtCheck
```

Plugins and projects that contribute additional lint-style checks (custom formatters, schema validators, license linters, etc.) want to fold their check into this same step so it runs alongside the existing ones — same matrix cell, same conditions, same "lint failures show up here" semantics. There's no first-class way to do that today; the only option is a fragile rewrite of `githubWorkflowBuild`:

```scala
ThisBuild / githubWorkflowBuild ~= {
  _.map {
    case step: WorkflowStep.Sbt if step.name == Some("Check headers and formatting") =>
      step.withCommands(step.commands :+ "myLintCheckAll")
    case other => other
  }
}
```

This is nine lines per plugin, repeats the magic step-name string at every call site, and breaks if upstream renames the step or splits it. It also doesn't compose: two plugins both rewriting `githubWorkflowBuild` need to be careful about order and not stomping each other.

## Proposed shape

A public setting that's just a list of extra commands appended into the existing lint step:

```scala
val tlCiLintCommands = settingKey[Seq[String]](
  "Additional sbt commands appended to the 'Check headers and formatting' step"
)

// existing default rendering becomes something like:
WorkflowStep.Sbt(
  List(
    s"$$ $${{ matrix.scala }}",
    "headerCheckAll",
    "scalafmtCheckAll",
    "project /",
    "scalafmtSbtCheck",
  ) ++ tlCiLintCommands.value,
  name = Some("Check headers and formatting"),
  cond = ...,
)
```

User code becomes one line:

```scala
ThisBuild / tlCiLintCommands += "smithyFmtCheckAll"
```

…and plugins can opt projects in by default:

```scala
override def projectSettings = Seq(
  tlCiLintCommands += "smithyFmtCheckAll",
)
```

`+=`'d sequences compose naturally across plugins and project settings, so there's no ordering footgun.

## Why it's worth doing

- Removes a copy-paste recipe from documentation. We currently tell `SmithyFormatPlugin` users (in [polyvariant/smithy-trait-codegen-scala](https://github.com/polyvariant/smithy-trait-codegen-scala)) to write the nine-line `githubWorkflowBuild ~= { … }` snippet by hand — every reader has to understand step matching, `WorkflowStep.Sbt`, and `withCommands` just to bolt one task onto CI.
- Plays nicely with autoplugins. A plugin that wants to opt projects into a check on CI can do so transparently via `projectSettings`, with users' explicit additions composing on top.
- The current pattern is brittle to upstream changes (step name, step splitting) — a named setting decouples user code from the rendering layout.

Happy to send a PR if the shape lands.

kubukoz:
Closes #888

### DIFF
--- a/ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
+++ b/ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
@@ -56,6 +56,9 @@ object TypelevelCiPlugin extends AutoPlugin {
       settingKey[String](
         "Condition for checking on CI whether this project is a fork of another (default: `github.event.repository.fork == false`)")
 
+    lazy val tlCiLintCommands = settingKey[Seq[String]](
+      "sbt commands run in the 'Check headers and formatting' CI step (default: derived from `tlCiHeaderCheck`, `tlCiScalafmtCheck`, and `tlCiJavafmtCheck`)")
+
   }
 
   import autoImport._
@@ -69,6 +72,19 @@ object TypelevelCiPlugin extends AutoPlugin {
     tlCiDocCheck := false,
     tlCiDependencyGraphJob := true,
     tlCiForkCondition := "github.event.repository.fork == false",
+    tlCiLintCommands := {
+      val headers = List("headerCheckAll").filter(_ => tlCiHeaderCheck.value)
+
+      val scalafmt = List(
+        "scalafmtCheckAll",
+        "project /",
+        "scalafmtSbtCheck"
+      ).filter(_ => tlCiScalafmtCheck.value)
+
+      val javafmt = List("javafmtCheckAll").filter(_ => tlCiJavafmtCheck.value)
+
+      headers ++ javafmt ++ scalafmt
+    },
     githubWorkflowTargetBranches ++= Seq(
       "!update/**", // ignore steward branches
       "!pr/**" // escape-hatch to disable ci on a branch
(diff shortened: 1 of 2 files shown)


ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
```diff
@@ -69,6 +72,19 @@ object TypelevelCiPlugin extends AutoPlugin {
     tlCiDocCheck := false,
     tlCiDependencyGraphJob := true,
     tlCiForkCondition := "github.event.repository.fork == false",
+    tlCiLintCommands := {
+      val headers = List("headerCheckAll").filter(_ => tlCiHeaderCheck.value)
```
kubukoz, line 76: This makes me wonder if `tlCiHeaderCheck` should just be deprecated in favor of it being the default here. But probably not

armanbilge: Hmm, wdym?

ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
```diff
@@ -69,6 +72,19 @@ object TypelevelCiPlugin extends AutoPlugin {
...
+
+      val scalafmt = List(
+        "scalafmtCheckAll",
+        "project /",
```
armanbilge, line 80: So, one thing to note, is that `project /` actually changes the state of the shell. So it makes a difference if commands are before or after this.

docs/customization.md
```diff
@@ -67,6 +67,7 @@ Both plugins are documented in [**sbt-typelevel-github-actions**](gha.md).
 - `tlCiDocCheck` (setting): Whether to build API docs in CI (default: `false`).
 - `tlCiDependencyGraphJob` (setting): Whether to add a job to submit dependencies to GH (default: `true`).
 - `tlCiForkCondition` (setting): Condition for checking on CI whether this project is a fork of another (default: `github.event.repository.fork == false`).
+- `tlCiLintCommands` (setting): sbt commands run in the "Check headers and formatting" CI step. Defaults are derived from `tlCiHeaderCheck`, `tlCiScalafmtCheck`, and `tlCiJavafmtCheck`. Append your own check (e.g. `tlCiLintCommands += "myLintCheckAll"`) to fold it into the same step. Scalafix is gated separately via `tlCiScalafixCheck` and runs as its own step.
```
armanbilge, line 70: Bikeshedding the name ... the idea is to separate "fast" commands (formatting, headers) from "slow" commands (Scalafix, which often requires compiling the code). But I'm not sure if "lint" fully captures that, since arguably Scalafix is also a form of linting.

armanbilge (commented):
Overall not against this, I think there are just some annoying gotchas to be aware of: state-changing commands and non-deterministic ordering (see https://github.com/http4s/sbt-http4s-org/pull/120#issuecomment-1008496880).

docs/customization.md
kubukoz: `Lint` -> `FastLint` would be fine with me :)

ci/src/main/scala/org/typelevel/sbt/TypelevelCiPlugin.scala
kubukoz: I mean that instead of the current state we could have a default value of `tlCiLintCommands := List("headerCheckAll", "scalafmtAll")`, and get rid of `tlCiHeaderCheck` as a key. If someone wants to opt out of header checks, they can modify `tlCiLintCommands`.

PR #908, opened 2026-07-15
Update `actions/upload-artifact` and `actions/download-artifact`

xuwei-k:


satorg:
Just to clarify: is it a routine version bump up, or can it help to resolve some issue? In the latter case, some context would be highly appreciated. Thank you!

mzuehlke:
I assume this is similar to #906 just a regular version bump.

